PR #1, opened 2011-03-18
PostgreSQL - serial pseudotype support

vaclavn:
SERIAL pseudotype is useful for quite every creation of table. It will be nice if Adminer will have support for it.


suphm:
>serial and bigserial are automatically created by checking the auto_increment radio.
not affected by alter table operation

PR #111, opened 2014-11-25
Modified for working like Editor

Zemistr:


Zemistr:
I'm sorry, but this is 2 years old thing. 😄 
I already changed this and I really don't have this piece of code.
We can close that and maybe one day in future someone else will do this. 😉

PR #112, opened 2014-12-03
PostgreSQL require host for connect.

h4kuna:
Default is localhost as placeholder in login form.


vrana:
I believe that the default value could be different. http://php.net/pg-connect#refsect1-function.pg-connect-parameters talks about default parameters but I don't know where the defaults are specified.

PR #116, opened 2014-12-07
List of tables implemented as <ul>

peterpp:
Encapsulation of list items is necessary for advanced CSS themes.

### DIFF
--- a/adminer/include/adminer.inc.php
+++ b/adminer/include/adminer.inc.php
@@ -884,17 +884,18 @@ bodyLoad('<?php echo (is_object($connection) ? substr($connection->server_info, 
 	* @return null
 	*/
 	function tablesPrint($tables) {
-		echo "<p id='tables' onmouseover='menuOver(this, event);' onmouseout='menuOut(this);'>\n";
+		echo "<ul id='tables' onmouseover='menuOver(this, event);' onmouseout='menuOut(this);'>\n";
 		foreach ($tables as $table => $status) {
-			echo '<a href="' . h(ME) . 'select=' . urlencode($table) . '"' . bold($_GET["select"] == $table || $_GET["edit"] == $table, "select") . ">" . lang('select') . "</a> ";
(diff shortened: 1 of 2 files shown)


stamster:
In Adminer, HTML table is not formatted how it should be.
There's a </td> closing tag missing for each row?!

neTpyceB:
Why a lot of merge requests are ignored?

djmattyg007:
The change to gitignore the .idea directory really shouldn't be in this repository. You should place it in your global gitignore file: https://thomashunter.name/blog/global-gitignore-vs-repository-gitignore/

> There's a closing tag missing for each row?!

Adminer aims to have a small a filesize as possible. One of the ways it does this is by abusing the fact that browsers will deal with these missing tags for legacy reasons.

> Why a lot of merge requests are ignored?

I am finding this very frustrating myself personally. I submitted eight pull requests almost three months ago and haven't heard back about any of them. I'm currently maintaining my own branch with all of my modifications mixed in: https://github.com/djmattyg007/adminer/tree/custom-v4.2.4

vrana:
I'm sorry, maintaining Adminer isn't much fun for me anymore. I personally use it only now and then.

peterpp:
@vrana, thanks for your reply. It's not optimistic, but it is much better than nothing. My pull request is more than year old and I lost all hopes for hear anything from you. I understand your point of view, but be sure there ale lots of people that use Adminer nearly every day. So my question is: what's your suggestion for those who want to help and improve this project?

@djmattyg007, thank you for feedback. I'll try to change my pull request.

peterpp:
I reduced this pull request to one minimalist commit.

djmattyg007:
@vrana what state do you think the project is in right now? If you were to hand the project over to someone else or bring someone else on board to help maintain it, what would you want them to know about it?

vrana:
Adminer is in maintenance mode right now. I release versions with security fixes, new translations and some random stuff. If you want more, I would suggest forking the project. Consider also renaming the fork so that it's not confusing to the users.

It's possible that I will start using Adminer more in July 2016. If that happens, the pace of development will probably increase.

vrana:
Merged. Please make sure that this doesn't break the designs (at least the top ones at https://www.adminer.org/#extras).

PR #117, opened 2014-12-17
Changed NULL format

newsiram:
- <i> was not italics
- reduce font & showed border

image


djmattyg007:
Your commit messages & pull request summary should probably mention that you've modified one specific theme.

PR #119, opened 2015-01-06
Fix tables list alphabetical order in PostgreSQL driver

finwe:
Returns tables list to alphabetical order which was broken in vrana:7a370e7


vrana:
This seems to be ordering normal tables first and then materialized views second. I guess it would be better to merge the results and order them together.

finwe:
Oh. Updated to always append the order by clause and rebased to master.

vrana:
Merged, thanks.

PR #120, opened 2015-01-06
Order sequences list by name

finwe:
Orders PostgreSQL sequences list by name rather than by default data table position


vrana:
Merged, thanks.

PR #124, opened 2015-02-26
Added format_bytes function

n0nag0n:
Sometimes the bytes get a little hard to read. This simple function makes it a little easier. This would probably work better as a plugin, but I didn't see a way to successfully do that :-\


vrana:
I like that I can see the biggest tables immediately. They don't stand out as much when using unit prefixes.

n0nag0n:
Fair enough, I can see that as a benefit. I'll just leave my local copy with the fix then.

PR #126, opened 2015-02-26
Format Data_length, Index_length, Data_free Totals

n0nag0n:


vrana:
I like that I can see the biggest tables immediately. They don't stand out as much when using unit prefixes.

PR #127, opened 2015-02-28
added character set utf8 to connection

Konrad90:
I discovered a bug. When using adminer on a strato.de shared host mysql database, it is not displaying field values containing äöüß-characters. Had to add this to make it work. Strato.de databases are created as latin1_german1_ci. Changing to utf8_general_ci doesn't fix the issue. 
Here's a config dumb of strato.de mysql server and newly created database: https://gist.github.com/Konrad90/a4f593059459b14af2aa 
I haven't figured out which setting is causing the issue.


Konrad90:
found the actuall issue https://github.com/vrana/adminer/pull/128

PR #128, opened 2015-02-28
set_charset was not getting defined on mysqli

Konrad90:
I discovered a bug. When using adminer on a strato.de shared host mysql database, it is not displaying field values containing äöüß-characters. Had to add this to make it work.

### DIFF
--- a/adminer/drivers/mysql.inc.php
+++ b/adminer/drivers/mysql.inc.php
@@ -39,6 +39,10 @@ if (!defined("DRIVER")) {
 			function quote($string) {
 				return "'" . $this->escape_string($string) . "'";
 			}
+
+            		function set_charset($charset) {
+                		$this->query("SET NAMES $charset");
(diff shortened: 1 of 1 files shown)


vrana:
There shouldn't be a reason to override [this method](http://php.net/mysqli.set-charset) as it also makes the client aware of the encoding (which affects e.g. escaping).

Konrad90:
I see. mysqli::set_charset('utf8') is working. 

mysqli::set_charset('utf8mb4') is not tough. 

$this->query("SET NAMES $charset") works with both.

vrana:
I think this happens when the MySQL client library used in PHP doesn't support utf8mb4. I've fixed it by using both ways, please see if it fixes your issue.

Konrad90:
yes. new mysql.inc.php fixed the issue on strato.de. thanks

PR #129, opened 2015-03-11
fix logger for SQL commands

mathroc:
Queries from SQL commands were not logged by sql-logs plugin

This PR add a pluggable sqlCommandQuery method to Adminer that is called by sql-logs (and could be used to customize query output)

### DIFF
--- a/adminer/sql.inc.php
+++ b/adminer/sql.inc.php
@@ -93,7 +93,7 @@ if (!$error && $_POST) {
 						$empty = false;
 						$q = substr($query, 0, $pos);
 						$commands++;
-						$print = "<pre id='sql-$commands'><code class='jush-$jush'>" . shorten_utf8(trim($q), 1000) . "</code></pre>\n";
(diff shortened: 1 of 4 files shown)


mathroc:
can this be merged ? or is there something to improve in this PR ?

mathroc:
I've updated the PR to fix the conflicts against master. anything preventing this from getting merged ?

adminer/sql.inc.php
```diff
@@ -94,7 +94,8 @@
 						$empty = false;
 						$q = substr($query, 0, $pos);
 						$commands++;
-						$print = "<pre id='sql-$commands'><code class='jush-$jush'>" . shorten_utf8(trim($q), 1000) . "</code></pre>\n";
+
+						$print = "<pre id='sql-$commands'><code class='jush-$jush'>" . $adminer->sqlCommandQuery($q, 1000) . "</code></pre>\n";
```
thaeriac: Why is there second parameter if function has only one?

mathroc: no clue ^^ I'm going to fix that

vrana:
Merged.

PR #133, opened 2015-06-22
Remove redundant dumpTable

Gargaj:
Is this necessary here? When exporting specific rows from a table, the table schema / creation string is never actually dumped with it - since the style is always empty string, the only thing it may potentially output is a `SET NAMES UTF8MB4`. 

Removing this would allow the export function to work on drivers (like MSSQL) that don't have a `create_sql()` function implemented.


vrana:
Many plugins need to hook here.

PR #134, opened 2015-06-24
Export table feature: Allow exporting of one query for each row

merijnvdk:
In some cases I would like to export table data with one query for each row in the table. This is useful for when you compare exported data or manually pick queries (rows). This was something I missed after switching from phpMyAdmin
- the dumpData function now checks for 'single_row_insert' param in $_POST
- added a checkbox in export interface
- added new language key 'One query for each row' and added it to nl (and xx) language

### DIFF
--- a/adminer/dump.inc.php
+++ b/adminer/dump.inc.php
@@ -157,7 +157,8 @@ echo "<tr><th>" . lang('Tables') . "<td>" . html_select('table_style', $table_st
 	. (support("trigger") ? checkbox("triggers", 1, $row["triggers"], lang('Triggers')) : "")
 ;
 
-echo "<tr><th>" . lang('Data') . "<td>" . html_select('data_style', $data_style, $row["data_style"]);
+echo "<tr><th>" . lang('Data') . "<td>" . html_select('data_style', $data_style, $row["data_style"])
+	. checkbox("single_row_insert", 1, $row["single_row_insert"], lang('One query for each row'));
 ?>
 </table>
 <p><input type="submit" value="<?php echo lang('Export'); ?>">
--- a/adminer/include/adminer.inc.php
+++ b/adminer/include/adminer.inc.php
@@ -671,6 +671,10 @@ focus(document.getElementById('username'));
 	function dumpData($table, $style, $query) {
 		global $connection, $jush;
 		$max_packet = ($jush == "sqlite" ? 0 : 1048576); // default, minimum is 1024
+		if ($_POST["single_row_insert"]) {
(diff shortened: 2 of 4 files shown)


helix84:
:+1: tested on MySQL

djmattyg007:
i'm pretty sure you need to do an `isset()` check on the `$_POST` parameter you're trying to access. If a checkbox isn't ticked, browsers won't send through any data for that field at all. You'll then try and access an array key that doesn't exist, causing a PHP warning.

merijnvdk:
true

djmattyg007:
Merging master into feature branches is generally considered a bad idea. You should rebase your branch on top of master first. In this case, you should obviously do this after you've reverted the merge.

djmattyg007:
I'd also recommend reading this: http://tbaggery.com/2008/04/19/a-note-about-git-commit-messages.html

To summarise, good commit messages should have a well-defined subject line with a hard limit of 72 characters, and a separate body.

merijnvdk:
Okay, I was struggling how to get my branch in sync with vrana's master (no git xp, just svn). Checking the rebase thing now, was put off by the 'use with caution' in the UI, but seems to make sense ;)

merijnvdk:
Okay, I rebased my local branch, but without the merge (and merge revert) to vranas master. Which basically means i've rewritten my history. 

Now my 'git push was rejected from remote repository the tip of your current branch is behind..' This is i assume because my public master has the merge which is no longer in my local branch.

I am unsure how I should solve this in the proper way. If I force the push then my public commit with the master merge will be gone. But that would be weird as we already discussed it here.
But when rebasing it would not include the merge, only the revert of the merge (which i did on my local master). Also it would be a strange commit which would have no changes because the rebase would have taken care of that..
I am a bit lost how to properly handle this.

helix84:
I recommend to force the push. GitHub supports that in the way that it will hide the old commit (before rebase) including any line-comments on it (there aren't any such comments here) and this pull request will only show the current commit(s).

merijnvdk:
Okay, I forced the push, hopefully its okay now

On a side note: there seem to be more similar issues with possible undefined arrey keys in the code (see also the other checkboxes on the dump page). But adminer seems to be suppressing notices and only shows warnings/errors (`error_reporting(6135)`).

helix84:
@merijnvdk looks good!

vrana:
I don't see a strong use case for this. Adminer tries to be minimalistic so every extra option must be really worth it.

Adminer mutes notices so there's no need for `isset` checks.

PR #135, opened 2015-06-26
PHP 7 compatibility fixes

jakoch:
Hey Jakub,

i've renamed the methods `Min_Result()`, `Min_DB()`, `Min_SQL()`, `TmpFile()` to `__construct()`.
This stops Adminer from throwing the warning: "Deprecated: Methods with the same name as their class will not be constructors in a future version of PHP."

Best, Jens


vrana:
Merged #142 instead.

PR #136, opened 2015-06-27
Protocol-relative linking from Google Fonts

jspetrak:


djmattyg007:
Best practice is now to always use HTTPS, even on HTTP pages. There really isn't a reason to use a protocol-independent URL any longer.

PR #137, opened 2015-07-09
PostgreSQL export

ejzy:
Basic export of PostgreSQL table structure - including columns, sequences, indexes, foreign keys, comments, trigger definitions (no trigger functions).
May not be perfect, but still enough for easy database managment, structure comparsion etc.


ejzy:
Replaced with #152 , closing...

PR #140, opened 2015-08-13
Add a responsive theme based on Nette

AMDG2:
This pull request add a responsive theme and and the viewport meta tag to enable responsiveness with the theme.

### DIFF
--- a/designs/nette-responsive/adminer.css
+++ b/designs/nette-responsive/adminer.css
@@ -1,0 +1,629 @@
+/*
+ * Theme "Nette-Responsive" for Adminer, (c) Baudouin Feildel
+ * This theme is based on the following:
+ *    Theme "Nette" for Adminer, (c) David Grudl
+ */
+
+/*
+it is based on some parts of:
+- Nette Framework design - http://api.nette.org
+- CSS by Brade - http://www.bradezone.com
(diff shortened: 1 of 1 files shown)


vrana:
I'm not sure that I want width=device-width,initial-scale=1 for Adminer. There's usually lots of data to display.

peterpp:
In my opinion, "width=device-width,initial-scale=1" is currently the best practice not only for responsive websites. This means that confortable clicking and readable font size have bigger priority than displaying whole page. If you want to see more data at once, you can still zoom the page.

AMDG2:
I agree with @peterpp I use a lot adminer from my mobile device. Because of that I really need a responsive interface. 
@vrana If you don't wan't to integrate this patch perhaps we can add a hook for extensions to add HTML tags in the header ?

peterpp:
@AMDG2, you can tweak meta headers in plugin. Take a look at our responsive theme: https://github.com/pematon/adminer-theme

AMDG2:
@peterpp Great theme, I did not knew about that one. 
@vrana Is it possible to add a link on adminer page to this theme ? So people who want to use adminer from mobile could easily know about this responsive theme !

vrana:
Implement this as a plugin. Linked https://github.com/pematon/adminer-theme from https://www.adminer.org/plugins/.

PR #141, opened 2015-08-14
Breadcumb no longer blocks the logout button.

0xACE:
Not sure why the top bar is blocking the logout button. This patch allows the user  to see and click the logout button.


djmattyg007:
Your commit message & pull request summary should probably explain that this is to do with one specific theme (and which theme it is). I clicked on this wondering why I'd never run into the issue myself, only to find it's because I'm not actually using the theme you've modified.

vrana:
This hid the breadcrumb completely for me. Added `.logout{z-index:3;}` instead.

PR #144, opened 2015-09-06
Display number of schemas in databases overview for PostgreSql

kamrandotpk:
Hi, please let me know if this commit is any useful or if any changes are required. Thanks.

### DIFF
--- a/adminer/include/adminer.inc.php
+++ b/adminer/include/adminer.inc.php
@@ -57,6 +57,13 @@ class Adminer {
 		return schemas();
 	}
 
+	/** Get list of schemas
+	 * @return array
+	 */
+	function count_schemas() {
+		return count_schemas();
+	}
+
 	/** Specify limit for waiting on some slow queries like DB list
 	* @return float number of seconds
 	*/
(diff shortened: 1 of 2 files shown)


adminer/include/adminer.inc.php
```diff
@@ -57,6 +57,13 @@ function schemas() {
 		return schemas();
 	}
 
+	/** Get list of schemas
```
vrana: It doesn't return list but count.

adminer/include/adminer.inc.php
```diff
@@ -762,6 +769,8 @@ function dumpHeaders($identifier, $multi_table = false) {
 	* @return bool whether to print default homepage
 	*/
 	function homepage() {
+		global $adminer;
+		echo (support("scheme") && $_GET['ns'] === '' ? "<p>" . lang('This database contains ' . $adminer->count_schemas() . ' schemas.') . "</p>\n" : "");
```
vrana: This is not translatable. Use `%d` instead. Add any translations to `lang/xx.inc.php`.

adminer/include/adminer.inc.php
```diff
@@ -57,6 +57,13 @@ function schemas() {
 		return schemas();
 	}
 
+	/** Get list of schemas
+	 * @return array
+	 */
+	function count_schemas() {
```
vrana: Why is this method here? I don't expect it to be customized.

adminer/include/adminer.inc.php
vrana: This is not minimalistic as Adminer is. What about `Schemas: 123` with 123 linking to the list of schemas?

PR #146, opened 2015-09-16
[mssql] Don't free statement

Gargaj:
Freeing the statement here will cause the statement to destruct too early when `$result = $connection->store_result();` is called in `sql.inc.php` (previous `$result` being overwritten triggers the destructor, but they hold the same `_result`).
As a consequence it will cause errors and will not show more than one table.


Gargaj:
I removed the line but I doubt it's fixable on the sql.inc.php level; sql.inc.php assumes results have separate handles that can be allocated and deallocated, whereas MSSQL has a single result handle that has to be pushed along to move between tables.

PR #147, opened 2015-09-19
[mssql] Add namespace to routine call

Gargaj:
MSSQL is picky about this, shouldn't affect the drivers that don't use namespaces.


vrana:
There's the `table` function exactly for this purposed. Changed the code to use it and merged.

PR #148, opened 2015-09-21
Do not set input type as numeric if field is an array

K0n24d:
Hello,

Just a really small fix regarding an issue i stumbled accross:
input type should not be numeric if the field is an array as you would input something like {1,2,3}

Regards,
K0n24d


vrana:
Merged.

PR #149, opened 2015-10-01
Foreign key command ends with semicolon

MilanPala:
Foreign key command now ends with semicolon.


vrana:
This is used on several places and I'm not sure that adding `;` wouldn't break them. Does this solve a particular issue?

PR #150, opened 2015-10-05
New skin 'flat'

isra00:
This is a simple skin based on a few alterations on adminer's original CSS.


vrana:
Merged and added to https://www.adminer.org/#extras.

PR #152, opened 2015-10-13
Ejzy postgre sql export

helix84:
A rebase of #137 on current master.

:+1: I tested it and it enables most of the Export functionality.

### DIFF
--- a/adminer/drivers/pgsql.inc.php
+++ b/adminer/drivers/pgsql.inc.php
@@ -3,6 +3,7 @@ $drivers["pgsql"] = "PostgreSQL";
 
 if (isset($_GET["pgsql"])) {
 	$possible_drivers = array("PgSQL", "PDO_PgSQL");
+    $database_dump_addon = '';
 	define("DRIVER", "pgsql");
 	if (extension_loaded("pgsql")) {
 		class Min_DB {
@@ -244,10 +245,10 @@ ORDER BY 1";
 
 	function table_status($name = "") {
 		$return = array();
-		foreach (get_rows("SELECT c.relname AS \"Name\", CASE c.relkind WHEN 'r' THEN 'table' ELSE 'view' END AS \"Engine\", pg_relation_size(c.oid) AS \"Data_length\", pg_total_relation_size(c.oid) - pg_relation_size(c.oid) AS \"Index_length\", obj_description(c.oid, 'pg_class') AS \"Comment\", c.relhasoids::int AS \"Oid\", c.reltuples as \"Rows\", n.nspname
+		foreach (get_rows("SELECT c.relname AS \"Name\", CASE c.relkind WHEN 'r' THEN 'table' WHEN 'mv' THEN 'materialized view' WHEN 'f' THEN 'foreign table' ELSE 'view' END AS \"Engine\", pg_relation_size(c.oid) AS \"Data_length\", pg_total_relation_size(c.oid) - pg_relation_size(c.oid) AS \"Index_length\", obj_description(c.oid, 'pg_class') AS \"Comment\", c.relhasoids::int AS \"Oid\", c.reltuples as \"Rows\", n.nspname
 FROM pg_class c
 JOIN pg_namespace n ON(n.nspname = current_schema() AND n.oid = c.relnamespace)
-WHERE relkind IN ('r','v')
+WHERE relkind IN ('r','v','mv','f')
 " . ($name != "" ? "AND relname = " . q($name) : "ORDER BY relname")
 		) as $row) { //! Index_length, Auto_increment
 			$return[$row["Name"]] = $row;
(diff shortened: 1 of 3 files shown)


helix84:
Not yet sure where to put this - here is code for exporting procedures:

SELECT 'CREATE FUNCTION ' || routine_name || '() RETURNS ' || data_type || ' AS $_$' ||
routine_definition || '$_$ LANGUAGE ' || external_language
FROM
    information_schema.routines 
WHERE
    specific_schema LIKE 'public'
    AND routine_name LIKE 'delete_address';

ejzy:
@helix84 This SQL looks nice&easy but does not work for functions with parameters because of missing parameters definition of the function. It's not really easy to get them and create SQL. I'll try to move forward with this, but the first step is to get pull-request merged.

ejzy:
Since 4.2.1 there is fatal error in exporting selected lines in PostgreSQL as the create_sql function is called (&missing). Merging this pull-request resolves this error as it adds "create_sql" function to pgsql driver.
Ping to @vrana, please consider merging this. Thank you.

helix84:
> Since 4.2.1 there is fatal error in exporting selected lines in PostgreSQL as the create_sql function is called (&missing).

Yes, that was my reason why I looked into this, too.

ejzy:
I made some preparations to export functions in pgsql, had to move some code to drivers. The function's parameters issue not resolved, bud the structure of code should be obvious.

https://github.com/ejzy/adminer/commit/6b173a7cd1989674b23d95273db18ce8abd5d059

helix84:
Perhaps this will help us with getting the arguments (I didn't dig into it yet):

http://git.postgresql.org/gitweb/?p=postgresql.git;a=blob;f=src/bin/pg_dump/pg_dump.c;h=b8eaa2a5404405e837ab84b2df8da89938077e67;hb=REL9_4_STABLE#l9774

helix84:
Here's my first attempt based on pg_dump's dumpFunc(). Needs testing.

```sql
SELECT 'CREATE FUNCTION ' || proname || '(' || funcargs || ') RETURNS ' || funcresult || E'\n' ||
'AS $$ ' || prosrc || ' $$ LANGUAGE ' || lanname || a.window || a.volatile || a.strict || a.secdef || leakproof || cost || rows
FROM (
  SELECT proname, proretset, prosrc, probin, 
    pg_catalog.pg_get_function_arguments(b.oid) AS funcargs, 
    pg_catalog.pg_get_function_result(b.oid) AS funcresult,
    proconfig,
    CASE proiswindow
      WHEN 't' THEN ' WINDOW'
      ELSE ''
    END AS window,
    CASE provolatile
      WHEN 's' THEN ' STATIC'
      WHEN 'i' THEN ' IMMUTABLE'
      ELSE ''
    END AS volatile,
    CASE proisstrict
      WHEN 't' THEN ' STRICT'
      ELSE ''
    END AS strict,
    CASE prosecdef
      WHEN 't' THEN ' SECURITY DEFINER'
      ELSE ''
    END AS secdef,
    CASE proleakproof
      WHEN 't' THEN ' LEAKPROOF'
      ELSE ''
    END AS leakproof,
    CASE
      WHEN (((SELECT lanname FROM pg_catalog.pg_language WHERE oid = prolang) IN ('internal', 'c')) AND procost != 0)
      THEN ' COST ' || procost
      ELSE ''
    END AS cost,
    CASE
      WHEN (proretset = 't' AND prorows != 0 AND prorows != 1000) THEN ' ROWS ' || prorows
      ELSE ''
    END AS rows,
    -- TODO proconfig
    (SELECT lanname FROM pg_catalog.pg_language WHERE oid = prolang) AS lanname
  FROM pg_catalog.pg_proc b
  WHERE proname = 'dup'
) AS a
```

ejzy:
Looks good. Should be changed to the current_schema() to list functions of selected schema, but it's just cosmetic. We should consider version of postgresql, because 'leakproof' is supported 9.2+, which is not so old version and could break compatibility for some users ( i would keep compatibility for major version). There are some other issues regarding 8.4- which is older major version and could be abandoned IMHO.

helix84:
> AND proname = 'FUNCNAME' AND nspname = 'public'

This is the function name and schema, to be parametrized.

I created this according to the latest supported pg version in pg_dump 9.4, but I agree we should make the newer functions optional depending on detected driver version. I'd also support 8.4, I still have one server to test with. Which problems do you see in 8.4? Seems to work for me (when I remove leakproof) and dumpFunc() also treats <8.4, 9.2) the same.

ejzy:
Sure, we can just place 

```
AND nspname = current_schema()
```

to list all functions of given schema. 

If you can test it on 8.4, let's keep it without leakproof for now and it should work on 8.4+, which should be enough.

ejzy:
As we mentioned the bug in exporting select results, #133 should solve it as well, without merging PostgreSQL export functionality into master.

jamgregory:
Has any progress been made on merging this into Adminer? This functionality would be really useful for us as we're struggling along with phpPgAdmin (which hasn't been updated in ~3 years)

smuuf:
How is this not merged yet? :disappointed:

ejzy:
I gave up, there is no progress. @vrana stated in other issue he is currently busy and only minor changes are merged.

vrana:
Merged the first commit. There were some conflicts and I fixed some style issues. Please test it.

About the second commit: resolve "needs more inspection", avoid a global variables, indent using tabs and create a separate pull request for it.

helix84:
Unfortunately, trying 4.2.6-dev, I get an error on Postgres in the list of tables screen. The tables are still shown correctly.
```
ERROR: column "oid" does not exist
LINE 1: ...ion_size(c.oid) AS "Data_length", pg_indexes_size(oid) AS "I...
HINT: There is a column named "oid" in table "c", but it cannot be referenced from this part of the query.
```
(Despite what the error says, I don't have a table called "c".)
The export doesn't work, either - the output is interrupted after comment, \connect, comment.
Haven't had time to dig deeper at the moment.

vrana:
Please try it at the current HEAD. I've merged a bad pull request introducing the `cid` error but fixed it later.

helix84:
Thanks, now it passed my smoke test. The error message is gone and a simple table export works.

PR #154, opened 2015-11-26
Specify filename when saving SQL command

bairdj:
When exporting an SQL command from Adminer, it saves the file as "sql.csv" or "sql.csv.gz". This can make it difficult to work with multiple exports, as they become "sql (1).csv", "sql (2).csv" and so on, making it difficult to work out what each one actually is.

This change conditionally shows a text field when the user selects 'save' or 'gzip' from export. The extension will be added by Adminer. If left blank, it uses the default 'sql'.

I haven't added this to the dump database and select table pages, as the default behaviour of using the database/table name is more appropriate in those situations.

### DIFF
--- a/adminer/sql.inc.php
+++ b/adminer/sql.inc.php
@@ -1,6 +1,6 @@
 <?php
 if (!$error && $_POST["export"]) {
-	dump_headers("sql");
+	dump_headers(!empty($_POST['filename']) ? $_POST['filename'] : 'sql');
 	$adminer->dumpTable("", "");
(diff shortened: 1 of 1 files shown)


bairdj:
It's not really unnecessary configuration though because it's optional to give it a name. It will still default to sql.csv if left blank. 

It'd be preferable to allow a meaningful file name rather than just a time stamp

vrana:
Adminer is minimalistic. Every extra configuration needs to deserve to be added.

PR #156, opened 2015-12-04
Improvements to Process List page

djmattyg007:
- Add auto-refresh capability
- Add ability to show full queries in table

### DIFF
--- a/adminer/include/design.inc.php
+++ b/adminer/include/design.inc.php
@@ -22,6 +22,9 @@ function page_header($title, $error = "", $breadcrumb = array(), $title2 = "") {
 <meta http-equiv="Content-Script-Type" content="text/javascript">
 <meta name="robots" content="noindex">
 <meta name="referrer" content="origin-when-crossorigin">
+<?php if (!$error && isset($_GET["processlist"], $_GET["autorefresh"])) { ?>
+<meta http-equiv="refresh" content="<?php echo intval($_GET["autorefresh"]); ?>">
+<?php } ?>
 <title><?php echo $title_page; ?></title>
 <link rel="stylesheet" type="text/css" href="../adminer/static/default.css">
 <script type="text/javascript" src="../adminer/static/functions.js"></script>
--- a/adminer/include/design.inc.php
+++ b/adminer/include/design.inc.php
@@ -22,9 +22,6 @@ function page_header($title, $error = "", $breadcrumb = array(), $title2 = "") {
 <meta http-equiv="Content-Script-Type" content="text/javascript">
 <meta name="robots" content="noindex">
 <meta name="referrer" content="origin-when-crossorigin">
-<?php if (!$error && isset($_GET["processlist"], $_GET["autorefresh"])) { ?>
-<meta http-equiv="refresh" content="<?php echo intval($_GET["autorefresh"]); ?>">
-<?php } ?>
 <title><?php echo $title_page; ?></title>
 <link rel="stylesheet" type="text/css" href="../adminer/static/default.css">
 <script type="text/javascript" src="../adminer/static/functions.js"></script>
--- a/adminer/processlist.inc.php
+++ b/adminer/processlist.inc.php
@@ -59,3 +59,16 @@ if (isset($_GET["showfull"])) {
 ?>
 <input type="hidden" name="token" value="<?php echo $token; ?>">
 </form>
+<?php if (isset($_GET["autorefresh"])) { ?>
+<a href="<?php echo remove_from_uri("autorefresh"); ?>"><?php echo lang('Stop auto refresh'); ?></a>
+<?php } else { ?>
+<form action="">
+<?php hidden_fields_get(); ?>
+<input type="hidden" value="" name="processlist">
+<?php if (isset($_GET["showfull"])) { ?>
+<input type="hidden" name="showfull" value="">
+<?php } ?>
+<input type="text" name="autorefresh" value="5">
+<input type="submit" type="submit" value="<?php echo lang('Auto refresh'); ?>">
+</form>
+<?php } ?>
--- a/adminer/processlist.inc.php
+++ b/adminer/processlist.inc.php
@@ -13,7 +13,7 @@ page_header(lang('Process list'), $error);
 ?>
 
 <form action="" method="post">
-<table cellspacing="0" onclick="tableClick(event);" ondblclick="tableClick(event, true);" class="nowrap checkable">
+<table cellspacing="0" onclick="tableClick(event);" ondblclick="tableClick(event, true);" class="<?php if (!isset($_GET["showfull"])){?>nowrap <?php } ?>checkable">
 <?php
 // HTML valid because there is always at least one process
 $i = -1;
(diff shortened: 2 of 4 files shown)


djmattyg007:

djmattyg007:
I have again rebased this pull request on top of the latest work on the master branch. I've also added another commit that stops the autorefresh from occurring if an error occurs, such as the user being logged out. This caused issues if (for example) a MySQL server was shut down while the autorefresh is active, as it would keep autorefreshing trying to log the user back in, and eventually hit the 30-minute lockout.

adminer/include/design.inc.php
```diff
@@ -22,6 +22,9 @@ function page_header($title, $error = "", $breadcrumb = array(), $title2 = "") {
 <meta http-equiv="Content-Script-Type" content="text/javascript">
 <meta name="robots" content="noindex">
 <meta name="referrer" content="origin-when-crossorigin">
+<?php if (!$error && isset($_GET["processlist"], $_GET["autorefresh"])) { ?>
```
vrana: Move to a `header` call in `processlist.inc.php`.

adminer/processlist.inc.php
```diff
@@ -13,7 +13,7 @@
 ?>
 
 <form action="" method="post">
-<table cellspacing="0" onclick="tableClick(event);" ondblclick="tableClick(event, true);" class="nowrap checkable">
+<table cellspacing="0" onclick="tableClick(event);" ondblclick="tableClick(event, true);" class="<?php if (!isset($_GET["showfull"])):?>nowrap <?php endif; ?>checkable">
```
vrana: Use `{}` instead of `endif`.

adminer/processlist.inc.php
```diff
@@ -36,7 +36,7 @@
 			($jush == "sql" && $key == "Info" && preg_match("~Query|Killed~", $row["Command"]) && $val != "") ||
 			($jush == "pgsql" && $key == "current_query" && $val != "<IDLE>") ||
 			($jush == "oracle" && $key == "sql_text" && $val != "")
-			? "<code class='jush-$jush'>" . shorten_utf8($val, 100, "</code>") . ' <a href="' . h(ME . ($row["db"] != "" ? "db=" . urlencode($row["db"]) . "&" : "") . "sql=" . urlencode($val)) . '">' . lang('Clone') . '</a>'
+			? "<code class='jush-$jush'>" . (isset($_GET["showfull"]) ? $val . "</code>" : shorten_utf8($val, 100, "</code>")) . ' <a href="' . h(ME . ($row["db"] != "" ? "db=" . urlencode($row["db"]) . "&" : "") . "sql=" . urlencode($val)) . '">' . lang('Clone') . '</a>'
```
vrana: Use `h($val)`.

adminer/processlist.inc.php
```diff
@@ -51,6 +51,24 @@
 	echo ($i + 1) . "/" . lang('%d in total', max_connections());
 	echo "<p><input type='submit' value='" . lang('Kill') . "'>\n";
 }
+if (isset($_GET["showfull"])) {
+	echo "<p><a href='" . h(ME . "processlist=") . "'>" . lang('Hide full') . "</a>";
```
vrana: Add these messages to `lang/xx.inc.php`.

adminer/processlist.inc.php
```diff
@@ -51,6 +51,24 @@
...
+} else {
+	echo "<p><a href='" . h(ME . "processlist=&showfull=") . "'>" . lang('Show full') . "</a>";
+}
 ?>
 <input type="hidden" name="token" value="<?php echo $token; ?>">
 </form>
+<?php if (isset($_GET["autorefresh"])) { ?>
+<a href="<?php echo remove_from_uri("autorefresh"); ?>"><?php echo lang('Stop auto refresh'); ?></a>
+<?php } else { ?>
+<form action="<?php echo ME; ?>" method="get">
```
vrana: Delete `<?php echo ME; ?>` (it's the default). `ME` should be wrapped into `h` if printed.
Delete ` method="get"` (it's the default).

adminer/processlist.inc.php
```diff
@@ -51,6 +51,24 @@
...
+<?php hidden_fields_get(); ?>
+<input type="hidden" value="" name="processlist">
```
vrana: `name` then `value`.

djmattyg007:
Auto-refresh is off when the page loads, so you have to manually turn it on. Presumably you're not trying to select rows while it's on because you've done that intentionally, so I don't think that's much of an issue.

It's very useful when you're trying to monitor what a database is doing (especially when it is under heavy load) as it saves you from having to manually refresh the page every few seconds.

djmattyg007:
Thanks for the feedback on the code, I'll make the changes and update the PR soon.

adminer/include/design.inc.php
djmattyg007: I'm not sure what you mean by this sorry. Are you saying that I should replace the `meta http-equiv` tag with a `Location` header?

vrana: `header("Refresh: " . intval($_GET["autorefresh"]))`

adminer/include/design.inc.php
```diff
@@ -22,6 +22,9 @@ function page_header($title, $error = "", $breadcrumb = array(), $title2 = "") {
...
+<meta http-equiv="refresh" content="<?php echo $_GET["autorefresh"]; ?>">
```
vrana: This has XSS. Be careful about escaping in cases like this.

vrana:
With quick auto-refresh, it would be very hard to kill a process. Does it make sense to keep the checked checkboxes checked (you'll need JavaScript to send the form on the background).

djmattyg007:
If a process is long-running and causing problems, you're probably going to have enough time to hit the "Stop autorefresh" button and then check the checkbox. If you click "Stop autorefresh" and the process has finished, it doesn't matter anymore. Therefore, I don't think (at least at this stage) that keeping track of which checkboxes are checked is necessary right now. It will introduce additional complexity that isn't strictly required.

I'm just about to make the switch from a meta http-equiv tag to the Refresh header, and then I should be able to update this PR.

djmattyg007:
I've rebased my branch on top of the latest commits on the master branch, squashed it all down to two commits, and made almost all of the changes you requested. The one change I haven't made is to switch to a `Refresh` HTTP header in the process list file.

~~While I think it would still be possible to use the `Refresh` header over the `meta http-equiv` tag, I don't believe it's reasonable to implement the refresh in any place other than the header (`adminer/include/design.inc.php`). This appears to be the only reliable place to determine whether or not an error occurred during the current request for any particular reason. While there is a global `$error` variable, it is not always filled out. Additionally...~~

While writing that, I realised that the reason you wanted it moved to the process list file is because that file would not be loaded unless there were no errors with the request, or there was an error but it did not prevent the actual process list from being rendered.

I'll make the necessary update now :+1:

djmattyg007:
Okay, I've pushed up that last change. Let me know if anything else needs adjusting or improving.

adminer/processlist.inc.php
```diff
@@ -51,6 +51,25 @@
...
+	echo "<p><a href='" . h(ME . "processlist=") . "'>" . lang('Hide full') . "</a>";
+} else {
+	echo "<p><a href='" . h(ME . "processlist=&showfull=") . "'>" . lang('Show full') . "</a>";
+}
 ?>
 <input type="hidden" name="token" value="<?php echo $token; ?>">
 </form>
+<?php if (isset($_GET["autorefresh"]) && !$error) { ?>
+<a href="<?php echo remove_from_uri("autorefresh"); ?>"><?php echo lang('Stop auto refresh'); ?></a>
+<?php header("Refresh: " . intval($_GET["autorefresh"]));
```
vrana: This is too late, you need to send it before any HTML output.

adminer/processlist.inc.php
```diff
@@ -51,6 +51,25 @@
...
+if (isset($_GET["showfull"])) {
+	echo "<p><a href='" . h(ME . "processlist=") . "'>" . lang('Hide full') . "</a>";
+} else {
+	echo "<p><a href='" . h(ME . "processlist=&showfull=") . "'>" . lang('Show full') . "</a>";
+}
 ?>
 <input type="hidden" name="token" value="<?php echo $token; ?>">
 </form>
+<?php if (isset($_GET["autorefresh"]) && !$error) { ?>
+<a href="<?php echo remove_from_uri("autorefresh"); ?>"><?php echo lang('Stop auto refresh'); ?></a>
```
vrana: This is still not in `xx.inc.php`. You can add all new messages there by running `lang.php xx`.

djmattyg007:
My apologies, this is what happens when I attempt to write code late at night. I've fixed up all the issues you reported.

vrana:
I've patched this in and fixed the coding style to match Adminer but then decided that I'm not interested in this change.

djmattyg007:
Thank you for considering this proposal and taking the time to review it. It's much appreciated.

PR #157, opened 2015-12-04
Add basic support for MySQL replication

djmattyg007:


djmattyg007:

djmattyg007:

adminer/drivers/mysql.inc.php
```diff
@@ -972,6 +972,20 @@ function show_status() {
 		return get_key_vals("SHOW STATUS");
 	}
 
+	/** Get replication status of master or slave
+	* @return array ($name => $value)
```
vrana: Add `@param`.

adminer/drivers/mysql.inc.php
```diff
@@ -972,6 +972,20 @@ function show_status() {
...
+	*/
+	function replication_status($type) {
+		return get_rows("SHOW $type STATUS");
+	}
+
+	/** Get list of slaves connected to the master
+	* @return array ($name => $value)
```
vrana: Add `@param`.

adminer/replication.inc.php
```diff
@@ -0,0 +1,24 @@
+<?php
+page_header(lang('Replication'));
+
+echo "<p><b>Master status</b>\n";
+$masterReplicationStatus = replication_status("MASTER");
```
vrana: Use `$master_replication_status` for variables.

adminer/replication.inc.php
```diff
@@ -0,0 +1,24 @@
+<?php
+page_header(lang('Replication'));
+
+echo "<p><b>Master status</b>\n";
```
vrana: Make translatable.

adminer/replication.inc.php
```diff
@@ -0,0 +1,24 @@
...
+$masterReplicationStatus = replication_status("MASTER");
+echo "<table cellspacing='0'>\n";
+foreach ($masterReplicationStatus[0] as $key => $val) {
+    echo "<tr>";
+    echo "<th><code>" . $key . "</code>";
```
vrana: Escape (use `h`).

adminer/replication.inc.php
```diff
@@ -0,0 +1,24 @@
...
+    echo "<td>" . nbsp($val);
+}
+echo "</table>\n";
+
+$slaveReplicationStatus = replication_status("SLAVE");
+if (!empty($slaveReplicationStatus)) {
+    echo "<p><b>Slave status</b>\n";
```
vrana: As above.

adminer/replication.inc.php
```diff
@@ -0,0 +1,24 @@
...
+$masterReplicationStatus = replication_status("MASTER");
+echo "<table cellspacing='0'>\n";
+foreach ($masterReplicationStatus[0] as $key => $val) {
+    echo "<tr>";
+    echo "<th><code>" . $key . "</code>";
+    echo "<td>" . nbsp($val);
+}
+echo "</table>\n";
+
+$slaveReplicationStatus = replication_status("SLAVE");
```
vrana: As above.

adminer/drivers/mysql.inc.php
```diff
@@ -972,6 +972,20 @@ function show_status() {
...
+	*/
+	function replication_connected_slaves() {
```
vrana: This is not used.

djmattyg007:
It's useful for the same reason that phpMyAdmin provides a dedicated interface for this. People don't always want to have to look up the specific commands to run from a manual, especially when they're under time pressure.

adminer/drivers/mysql.inc.php
djmattyg007: This function doesn't have any parameters.

djmattyg007:
I've rebased this branch against the master branch and made all of the changes you requested in the hope that you reconsider this for inclusion. It is not possible to add this sort of thing with a plugin, and there is no reason why replication support could not be added for additional database systems. If I were to implement support for other systems, would you be more likely to reconsider it?

adminer/replication.inc.php
```diff
@@ -0,0 +1,24 @@
...
+if (!empty($slaveReplicationStatus)) {
+    echo "<p><b>Slave status</b>\n";
+    echo "<table cellspacing='0'>\n";
+    foreach ($slaveReplicationStatus[0] as $key => $val) {
+        echo "<tr>";
+        echo "<th><code class='jush-" . $jush . "status'>" . h($key) . "</code>";
```
vrana: What does the class mean?

djmattyg007: It's been a long time since I implemented this, and I don't remember exactly. As the `jush` system is used for syntax highlighting (including linkifying various SQL keywords), I assume I used it to ensure the various keywords (contained within `$key`) were linkified.

djmattyg007:
I have pushed up my changes to my branch, but it would appear that as long as this PR is closed, Github won't update the patchset associated with this PR. If you re-open it, it may work as expected. If it does not, I'm happy to open another PR as necessary.

adminer/replication.inc.php
vrana: It links to http://dev.mysql.com/doc/mysql/en/server-status-variables.html which doesn't document anything from http://dev.mysql.com/doc/mysql/en/show-slave-status.html.

vrana:
Please create a new pull request. This couldn't be reopened for this reason: "The replication branch was force-pushed or recreated." Make these changes:

- Link to the correct doc (using `doc_link`).
- Use `result` instead of `get_rows`.

I agree that this might be potentially useful.

vrana:
Oh and please attach a screenshot of the final version, I don't have replication.

djmattyg007:
I'm working on making the necessary modifications. Could I ask why `process_list()` uses `get_rows()`, but it isn't appropriate for `replication_status()`?

djmattyg007:
Having spent some time reviewing everything, I don't believe using `result()` is the right choice. It returns one field from one row, rather than the whole row.

vrana:
You are right, I got confused, sorry. We don't have a good abstraction for this. There's `get_key_vals` which is useful for tables with 2 columns and N rows. This result have N columns and 1 row.

djmattyg007:
No problem. I also wanted to note that I've implemented `doc_link` support, but unfortunately this page does not have anchor points for each individual key: https://dev.mysql.com/doc/refman/5.7/en/show-slave-status.html so I wasn't able to include support for that.

I should be ready to submit a new PR fairly soon.

vrana:
The solution for missing MySQL doc anchor links is to add them :-). I've done that before: https://bugs.mysql.com/bug.php?id=60016, https://bugs.mysql.com/bug.php?id=68989. Before they are added to the MySQL doc, you can link them from Adminer blindly.

djmattyg007:
You are quite right, thank you for the inspiration. I submitted the new PR right before reading this message, so I'll push an update in a moment that adds the anchor links.

PR #158, opened 2015-12-05
Add regex support to table filter plugin

djmattyg007:


djmattyg007:
I've rebased this pull request on top of the latest commits on the master branch.

djmattyg007:
Pull request #180 appears to be much more suitable for the job, especially since it stores the current search term in the user's `sessionStorage` so it persists across page loads. Barring any objections, I now intend to close this PR.

PR #159, opened 2015-12-05
Allow passing in of config filename to login server plugin

djmattyg007:
This means you can avoid editing a PHP file (and even store the list of
servers completely away from adminer.php if desired). Avoiding editing a
PHP file is desirable, as it could necessitate clearing an opcode cache
for the change to take effect.

### DIFF
--- a/plugins/login-servers.php
+++ b/plugins/login-servers.php
@@ -11,11 +11,19 @@ class AdminerLoginServers {
 	var $servers, $driver;
 	
 	/** Set supported servers
-	* @param array array($domain) or array($domain => $description) or array($category => array())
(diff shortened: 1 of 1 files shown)


djmattyg007:

djmattyg007:

vrana:
I don't see why this have to be done inside the plugin and not in the code calling this plugin.

djmattyg007:
It's useful when running this inside a container, where you want to provide a single distributable image where people don't touch anything inside the container and just configure everything with a few environment variables.

vrana:
The proposed code in the constructor could be in the code instantiating the plugin.

PR #160, opened 2015-12-05
Strip quotes from integers when querying mysql int datatype fields

djmattyg007:
If MySQL doesn't have to convert a string to an integer itself,
complicated queries run much faster. This is important for databases
under heavy load.


djmattyg007:

djmattyg007:

PR #161, opened 2015-12-06
Add plugin for printing table indexes structure in expanded format

djmattyg007:


djmattyg007:
I've rebased this commit on top of the latest work on the master branch.

djmattyg007:

vrana:
Merged with small changes:

1. Fixed authorship of the plugins.
1. Used `<th>` instead of `<td>` for headings.
1. Used PRIMARY instead of PRI.
1. Changed `<i>None</i>` to `&nbsp;`. It would need to be translated.
1. Inlined `support("comment")`. The compiler can remove the code of unsupported features but only with the direct use of `support`.

PR #162, opened 2015-12-06
Fix bug with index generation for new elements in selectAddRow()

djmattyg007:
Instead of incrementing the indice, the code would simply append the
string "1" over and over. It will now behave as expected, the with
indice incrementing by 1 each time.

### DIFF
--- a/adminer/static/functions.js
+++ b/adminer/static/functions.js
@@ -302,14 +302,17 @@ function selectAddRow(field) {
 	};
 	field.onchange();
 	var row = cloneNode(field.parentNode);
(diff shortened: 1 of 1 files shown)


djmattyg007:
I've rebased this pull request on top of the latest changes on the master branch.

djmattyg007:

vrana:
What's wrong with appending 1? The goal is to get a higher number which this does.

djmattyg007:
I figured it should be `thing1`, `thing2`, etc. Not `thing1`, `thing11`.

PR #155, opened 2015-12-04
fix Min_DB::connect (php7 compatibility)

B4nan:
Same purpose as #135, but without other changes

PR #164, opened 2015-12-09
Fix PHP7 Compatibility

iggyvolz:
See http://sourceforge.net/p/adminer/bugs-and-features/477/ for a description of the bug.

Using $_, $__, and $___ for unused parameters (which must be unique) - this can be anything as long as it is unused.

### DIFF
--- a/adminer/drivers/elastic.inc.php
+++ b/adminer/drivers/elastic.inc.php
@@ -58,7 +58,7 @@ if (isset($_GET["elastic"])) {
 				return $this->rootQuery(($this->_db != "" ? "$this->_db/" : "/") . ltrim($path, '/'), $content, $method);
 			}
 
-			function connect($server, $username, $password) {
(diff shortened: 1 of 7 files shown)


thaeriac:
:+1:

integer:
Same issue on my machine with PHP 7.0.1 and Adminer 4.2.3. 

@iggyvolz , thanks for your pull. Here is my small review: would be nice if you could change formatting (spaces) for default values as it is in rest of project. Unused parameters naming would be IMHO better as it is in `mysqli::connect` - `$database = null, $port = null, $socket = null`. Variables with underscores are little confusing.

integer:
@iggyvolz , did you see https://github.com/vrana/adminer/pull/155 - fixing same issue and keeps support for `$server` in format `<host>:<port>`. It's backward compatible.

iggyvolz:
Understood.  I thought that didn't fix this issue which is why I kept it open, but it looks like it does.

PR #165, opened 2015-12-18
Add HTTPS compatibility

Angristan:
Adding //url/ instead of http://url/ adds comptaibilty for HTTPS


gregtyler:
This should explicitly use "https://" rather than "//", since the HTTPS version of Google fonts is always available and a more secure connection.

Explanation and further links available on [Paul Irish's blog](http://www.paulirish.com/2010/the-protocol-relative-url/).

Angristan:
Yes sure.

Angristan:
https://github.com/vrana/adminer/pull/169

PR #166, opened 2015-12-19
Update adminer.css

Angristan:
I think it's better like this.


vrana:
Please talk with pappu687 before doing this change.

PR #168, opened 2016-01-08
Oracle driver fixes for a more general usage

meob:
Within the get_databases() function users granted to SELECT data by then table owner were not able to see the table structure since they do not have an entry in the USER_TABLESPACES view. Using the ALL_TABLES view gives the possibility to every user granted.

Generally only DBAs can see the DBA_SEGMENTS view which was used in the driver schema() function.
Using an easy filter in the ALL_TABLES view all "normal" users can obtain the full schema list.

### DIFF
--- a/adminer/drivers/oracle.inc.php
+++ b/adminer/drivers/oracle.inc.php
@@ -165,7 +165,7 @@ if (isset($_GET["oracle"])) {
 	}
 
 	function get_databases() {
-		return get_vals("SELECT tablespace_name FROM user_tablespaces");
+		return get_vals("SELECT distinct tablespace_name FROM all_tables where tablespace_name is not null");
 	}
 
 	function limit($query, $where, $limit, $offset = 0, $separator = " ") {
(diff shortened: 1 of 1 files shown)


adminer/drivers/oracle.inc.php
```diff
@@ -165,7 +165,7 @@ function connect() {
 	}
 
 	function get_databases() {
-		return get_vals("SELECT tablespace_name FROM user_tablespaces");
+		return get_vals("SELECT distinct tablespace_name FROM all_tables where tablespace_name is not null");
```
vrana: Why `DISTINCT`? What are the other columns of this table in rows with the same `tablespace_name`? Shouldn't you add a condition instead?

vrana: Use uppercase for keywords.

meob:
Hi,

the first query use a view that contains the list off all tablespace 
visible to the user
even if they do not contain any useful table.

The proposed query avoids unuseful tablespaces query the ALL_TABLES view.
In the ALL_TABLES view the same tablespace is repeaded several times...
to avoid duplicates DISTINCT or GROUP BY should be used.

I hope it is useful

meo

vrana:
Please provide an example of the data where the value of `table_space` is repeated.

table_space | column B <- what's this?
------------ | ----------------
table_a | value1 <- what's this?
table_a | value2 <- what's this?

PR #171, opened 2016-02-25
update select_db function to work with latest (MS SQL Azure) databases

mrbasil:
USE statement is no longer supported.  Instead reconnect using the specified database.  The change was only added to the sqlsrv driver as mssql is deprecated.

### DIFF
--- a/adminer/drivers/mssql.inc.php
+++ b/adminer/drivers/mssql.inc.php
@@ -39,7 +39,18 @@ if (isset($_GET["mssql"])) {
 			}
 
 			function select_db($database) {
-				return $this->query("USE " . idf_escape($database));
+			    	// 20160224 MB - to support SQL Azure re-connect instead of using deprecated USE statement
+				//return $this->query("USE " . idf_escape($database));
+				global $adminer;
+				$credentials = $adminer->credentials();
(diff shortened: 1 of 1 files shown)


adminer/drivers/mssql.inc.php
```diff
@@ -39,7 +39,18 @@ function quote($string) {
 			}
 
 			function select_db($database) {
-				return $this->query("USE " . idf_escape($database));
+			    	// 20160224 MB - to support SQL Azure re-connect instead of using deprecated USE statement
```
vrana: Delete.

adminer/drivers/mssql.inc.php
```diff
@@ -39,7 +39,18 @@ function quote($string) {
...
+				//return $this->query("USE " . idf_escape($database));
```
vrana: Delete.

adminer/drivers/mssql.inc.php
```diff
@@ -39,7 +39,18 @@ function quote($string) {
...
+				global $adminer;
+				$credentials = $adminer->credentials();
+				$this->_link = @sqlsrv_connect($credentials[0], array("UID" => $credentials[1], "PWD" => $credentials[2], "Database" => $database, "CharacterSet" => "UTF-8"));
```
vrana: Extract the code common with `connect()` to a method.

PR #172, opened 2016-02-28
Fix http:// links in stylesheets to protocol-independent ones (//)

piit79:
The http links do not load on https pages due to mixed content restriction
of modern browsers


vrana:
Merged, thanks.

PR #174, opened 2016-03-16
Elastica table_status() fix

IonBazan:
selectLinks were broken while connecting to Elastica. This patch should fix that.


vrana:
Merged.

PR #176, opened 2016-04-26
Improve limit check in edit-foreign plugin

Amunak:
The limit is enforced in the query to prevent out-of-memory errors for big target tables.


vrana:
Merged.

PR #177, opened 2016-05-11
PgSQL: scan SQL for COPY FROM stdin

finwe:
It is not possible to use `COPY FROM stdin` in web drivers. Such SQL queries result in hanging db/server/fpm-pool processes.

Scan SQL for these queries and throw an error when found.


vrana:
1. This is too specific. The query could be structured differently.
1. What if the matched pattern is inside a string or a comment?
1. I don't think that Adminer is the right layer to address this.
1. Use `lang()` for messages.

PR #179, opened 2016-05-31
Add javascript move for editing

jnvsor:
Adds a js onclick handler so you don't have to wait for several page loads for every reorder

### DIFF
--- a/adminer/include/editing.inc.php
+++ b/adminer/include/editing.inc.php
@@ -277,8 +277,8 @@ echo checkbox("fields[$i][has_default]", 1, $field["has_default"], "", "", "", "
 		echo "<td>";
 		echo (support("move_col") ?
 			"<input type='image' class='icon' name='add[$i]' src='../adminer/static/plus.gif' alt='+' title='" . lang('Add next') . "' onclick='return !editingAddRow(this, 1);'>&nbsp;"
-			. "<input type='image' class='icon' name='up[$i]' src='../adminer/static/up.gif' alt='^' title='" . lang('Move up') . "'>&nbsp;"
-			. "<input type='image' class='icon' name='down[$i]' src='../adminer/static/down.gif' alt='v' title='" . lang('Move down') . "'>&nbsp;"
+			. "<input type='image' class='icon' name='up[$i]' src='../adminer/static/up.gif' alt='^' title='" . lang('Move up') . "' onclick='return !editingMoveRow(this, 1);'>&nbsp;"
+			. "<input type='image' class='icon' name='down[$i]' src='../adminer/static/down.gif' alt='v' title='" . lang('Move down') . "' onclick='return !editingMoveRow(this, 0);'>&nbsp;"
 		: "");
 		echo ($orig == "" || support("drop_col") ? "<input type='image' class='icon' name='drop_col[$i]' src='../adminer/static/cross.gif' alt='x' title='" . lang('Remove') . "' onclick=\"return !editingRemoveRow(this, 'fields\$1[field]');\">" : "");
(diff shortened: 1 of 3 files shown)


adminer/include/editing.inc.php
```diff
@@ -290,7 +290,6 @@ function edit_fields($fields, $collations, $type = "TABLE", $foreign_keys = arra
 * @return bool
 */
 function process_fields(&$fields) {
-	ksort($fields);
```
vrana: Theoretically, browsers could send the fields in a random order. That's why the code was structured this way. I'm not sure if it's worth maintaining this assumption.

adminer/static/editing.js
```diff
@@ -283,6 +283,30 @@ function editingRemoveRow(button, name) {
 	return true;
 }
 
+/** Move table row for field
+* @param HTMLInputElement
+* @param bool direction
```
vrana: There's `boolean` in JavaScript, not `bool`.

adminer/static/editing.js
```diff
@@ -283,6 +283,30 @@ function editingRemoveRow(button, name) {
...
+* @return boolean
+*/
+function editingMoveRow(button, dir){
+	var row = parentTag(button, 'tr');
+	var table = row.parentNode;
+	var tags = table.getElementsByTagName('tr');
+	for (var i=0; i < tags.length; i++) {
```
vrana: Can you do this using `nextSibling`/`previousSibling` instead?

adminer/include/editing.inc.php
jnvsor: Unfortunately if you ksort the fields after reordering with javascript it'll put them right back where they were. An alternative would be to add a hidden "Position" input and tweak that with javascript, then sort based on that. Should I write that up?

adminer/static/editing.js
jnvsor: I'd have to use `nextElementSibling`/`previousElementSibling` which would drop support for IE8 and below - is that ok?

adminer/include/editing.inc.php
vrana: Thanks for the research, I'm glad that we could lift this limitation. I don't remember what made me think that the order could be random, maybe some obscure browser.

adminer/static/editing.js
vrana: Why can't you use `nextSibling`?

adminer/static/editing.js
```diff
@@ -271,6 +271,25 @@ function editingRemoveRow(button, name) {
...
 }
 
+/** Move table row for field
+* @param HTMLInputElement
+* @param boolean Direction to move row. True for up false for down
+* @return boolean Success
+*/
+function editingMoveRow(button, dir){
+	var row = parentTag(button, 'tr');
+
```
vrana: Detect if this feature is available in the browser and return `false` if not.

adminer/static/editing.js
jnvsor: Because text nodes are dom elements too (Including the whitespace between TR tags)

If we could ensure the TRs and the TBODY are always rendered without whitespace between them we could use nextSibling no problem, otherwise it might seem like the buttons are broken when it only reorders no-op whitespace

adminer/static/editing.js
jnvsor: > Detect

Detect `nextElementSibling` or detect `parentTag`? I believe `parentTag` is used without checks all over the place

vrana: Detect `nextElementSibling`.

vrana:
Merged with some changes.

PR #180, opened 2016-05-31
Update AdminerTablesFilter

jnvsor:
- Removes children request (Should work on IE8 and IE7 now)
- Uses `<strong>` instead of `<b>`
- Doesn't leave said `<strong>` tags behind after updating the list
- Highlights multiple matches in a single table name
- Works case insensitively
- Persists on page change
- Improves performance with setTimeout
    With 400 tables, the old implementation locks up the tab (or
    browser if using something without multiprocess) for about
    half a second per keyup. Yes, this includes modifiers that
    don't actually change the filter. The new version handles
    the same event in 0.09 milliseconds.
  
    That's with all the above improvements.
    Tested in firefox 45.1.1 performance monitor.

### DIFF
--- a/plugins/tables-filter.php
+++ b/plugins/tables-filter.php
@@ -7,7 +7,7 @@
 */
 class AdminerTablesFilter {
 	function tablesPrint($tables) { ?>
-<p class="jsonly"><input id="filter-field" onkeyup="tablesFilterInput();">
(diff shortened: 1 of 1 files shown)


vrana:
Merged.

jnvsor:
This isn't actually merged yet - or are you merging locally?

PR #181, opened 2016-06-27
Pg locks

soleuu:
Adds support for locks for postgresql


vrana:
This seems very similar to process list. Can you merge these two?

soleuu:
The idea is to make a similar view as with phppgadmin (project we were using for years not maintained anymore) with the kill support.

On a production system, the will be lots of locks, and lots of lock per query/transaction. Merging this with the process lists will make process lists complicated I think. 
This view allow us to have a global view of all locks on our system. And is used during emergency. As emergency didn't occur very often :
- you don't always remember how to see locks, 
- you don't always remember how to kill the request as fast as possible.

Our Context :
We use this view to detect when some request are locking data for a long time, to kill the blocking request.
One regular problem arise when one request need a long "accesShareLock". One request need one "exclusiveLock". All new new accessShareLock are then waiting the "accessExclusive" lock to be satisfied, making lots of connections on the system.



Note : having lots of locks is not a problem but, having lots of not granted locks is a problem.

vrana:
You've convinced me that this deserves a separate page. Merge it on the source code level then. Look at `variables=` and `status=` about how you can do it. This file is basically a copy/paste of `processlist.inc.php` with lots of irrelevant code.

PR #182, opened 2016-06-28
Added support for binary foreign keys selection on edit

alle:
Binary foreign keys can now be properly selected: an options list with empty values was generated.


vrana:
Merged.

PR #183, opened 2016-06-29
Checkbox plugin for bools in psql

kusmierz:
Real checkbox for bool in psql. It should also resolves [#483](https://sourceforge.net/p/adminer/bugs-and-features/483/).


vrana:
1. This should probably in the main code, not in a plugin.
1. Use tabs for indenting.

kusmierz:
Done.

vrana:
Always make sure that the code actually works. The function is supposed to print the data, not return it. Also look at the nearby code, the `name` attribute is already in `$attrs`, there's no need to build it again. Also preserve compatibility with PHP 5.3 (by not using short array syntax).

Merged with these changes.

kusmierz:
Huh, sorry Jakub - I made those changes late, and I missed I committing wrong version of my changes. Thank you for fixes and merge!

PR #184, opened 2016-06-29
Plugin: Use <select> on custom user types in psql

kusmierz:
It lists custom user types (`types()`), get them possible values (psql query) and shows as <select> list.


vrana:
Merged.

PR #187, opened 2016-07-07
JSON export formatting improvements for tabs and numeric values

jannecederberg:
Suggesting changes to `json_row` function in `adminer/include/functions.inc.php` for
1. Handling tab characters in database table values by converting them to \t so that  resulting JSON data validates and
2. Exporting integers and floats as numeric values instead of strings


vrana:
Server-side number precision might be higher than the client-side number precision.
Merged the `\t` change.

jannecederberg:
Ok, cool, thanks!

PR #188, opened 2016-07-15
foreign key link fix - added missing ns

honza889:
When I have foreign key defined across schemes (namespaces) in PostgreSQL, links on values in SELECT are invalid - keeps ns in URL unchanged.

This fix ensure the ns in link is correct - by foreign key target.


vrana:
1. What if there's no ns in the current URL but the foreign key has it? Is this possible?
2. Put the conditional statement body in `{}` and on a separate line (see nearby code).

honza889:
Updated by comment.

vrana:
What if there's no ns in the current URL but the foreign key has it? Is this possible?

honza889:
Ok, even through there is redirect when ns is not defined (to default ns=public), added condition. Maybe will be useful for same database, where is ns optional...

vrana:
Merged.

PR #189, opened 2016-08-18
Fix index length with Postgresql

PiR-B-:


vrana:
Merged.

vrana:
Reverted, this doesn't work in PostgreSQL < 9.0. What does the "fix" should do anyway? What doesn't work with the original code?

PR #190, opened 2016-08-30
postgresql error with index on hidden columns (ctid)

soleuu:
context:
On postgesql, if you have indexed hidden columns like ctid, there is some JS error on the json generation.

in order to test, add index : 

``` sql
create index on "SomeTable" (ctid);
```


vrana:
Merged.

PR #192, opened 2016-09-21
Changed 'geraakt' to 'beinvloed' and fixed capitalisation in 'Kies'

megaxlr:
A little fix in the Dutch language.

Beinvloed is more fitting in this context.


vrana:
Merged. 'kies' capitalization is correct, there's another translation for 'Select'. I've also changed the text to use UTF-8.

PR #194, opened 2016-10-10
Added pdo_dblib as a driver for MS-SQL

charlesqwu:
Added pdo_dblib as a driver for MS-SQL -- this is particularly useful for PHP7 as currently there is no native MS-SQL driver in PHP7.

### DIFF
--- a/adminer/drivers/mssql.inc.php
+++ b/adminer/drivers/mssql.inc.php
@@ -8,7 +8,7 @@
 $drivers["mssql"] = "MS SQL";
 
 if (isset($_GET["mssql"])) {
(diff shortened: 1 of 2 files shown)


TwanoO67:
I tried that one, but I'm stuck having those king of error on connection:
"SQLSTATE[HY000] General SQL Server error: Check messages from the SQL Server (severity 11)"

maxxer:
Tried the patch and it works!!! Please merge, newer PHP versions use dblib (at least on linux/macos)

vrana:
Merged.

PR #195, opened 2016-11-11
Change 'undefined' to 'null'

shaununderwood:
There's no undefined value in json. This update changes undefined to null.


shaununderwood:
you know, you should pull this in. ;)

vrana:
Merged. The `undefined` was used also in `null`, I've fixed it for consistency.

PR #197, opened 2016-11-11
added mongodb PHP7 beta driver.

pivolan:
Work only insert, update, list database, list index, list tables. select, where, limit, order, delete, edit, edit from select.
does not work structure modify, index, modify, custom queries, export, import, groups, collections drop, add.

### DIFF
--- a/adminer/drivers/mongodb.inc.php
+++ b/adminer/drivers/mongodb.inc.php
@@ -9,7 +9,7 @@ use MongoDB\Driver\Query;
 $drivers["mongodb"] = "MongoDB PHP7 (beta)";
 
 if (isset($_GET["mongodb"])) {
-	$possible_drivers = ["mongodb"];
+	$possible_drivers = array("mongodb");
 	define("DRIVER", "mongodb");
 
 	if (class_exists('MongoDb\Driver\Manager')) {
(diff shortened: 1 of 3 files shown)


vrana:
1. Please make sure that this compiles in PHP 5.3 (no short array syntax, maybe others).
1. Do not put `{` on the beginning of the line. Put it on the end of the previous line.

vrana:
There's already `mongo.inc.php`. Can you merge these two?

pivolan:
1. Fixed
2. fixed
In mongo.inc.php another version of driver, and there is nothing to merge. Versions between two drivers incompatible.

vrana:
See the other drivers. They are like:

    if (extension_loaded("mysqli")) {
      // mysqli specific code
    } elseif (extension_loaded("mysql") {
      // mysql specific code
    }

Do the same.

pivolan:
too many diffs, can't do this, sorry.

fidraj:
So this was never implemented?

pivolan:
I think yes. In old mongo drivers less functions than in new. And I don't know functionality in old mongodb drivers. So it's too difficult for me, to merge two versions.
Also I think that we shouldn't merge two drivers, because new client incompatible with old server, and old client incompatible with new client. So for different servers we should use difference drivers.
Sorry for my English.

PR #199, opened 2016-12-10
Optional config file for permanent login

peter-trerotola:
I connect to a ton of DBs all the time and switch computers and browsers quite often and lose all my permanent connections. I put this together so that I can take a config file and move it around with me to ensure my permanent connections can be recreated quickly. I tested this compiled and uncompiled and it was working well for me if anyone is interested.

### DIFF
--- a/adminer/include/auth.inc.php
+++ b/adminer/include/auth.inc.php
@@ -8,6 +8,44 @@ if (!$has_token) {
 $token = get_token(); ///< @var string CSRF protection
 
 $permanent = array();
+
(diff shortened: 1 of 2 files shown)


scrobbleme:
@peter-trerotola Nice addition.
Would be great, if at least the passwords could be encrypted using a master key, which has to be inserted everytime some one tries to switch the connection.

vrana:
This could be done using [customization](https://www.adminer.org/en/extension/), there's no need to create another abstraction.

PR #200, opened 2016-12-20
PHP Warning hidden and better parsing times (at least for postgresql)

kusmierz:
There was PHP Warning about missing `$field`'s key. And I changed `timeFormat` to better handle psql timestampes. Should be tested on other databases.


vrana:
Merged in #201.

PR #202, opened 2017-01-07
Different table size format

jmarcher:


vrana:
Please try to implement this as a plugin. Also support PHP 5.

PR #203, opened 2017-01-16
postgresql : ERROR:  function connection_id() does not exist

soleuu:
This pull request resolve the following error on postgresql database : 

ERROR:  function connection_id() does not exist at character 8
HINT:  No function matches the given name and argument types. You might need to add explicit type casts.
STATEMENT:  SELECT CONNECTION_ID()

Cause : specific mysql code to kill slow queries.


vrana:
Merged.

PR #204, opened 2017-01-17
postgresql : unique partial index musn't be considered as unique.

soleuu:
Partial unique index are considered as unique and cand can cause trouble on row edition : 
Example update multiples row at once when you want to update only one row.

I didn't concatenate partial with PRIMARY or UNIQUE because the function 'unique_array()' use one regular expression that isn't anchored with the begining nor the end of the string "$index['type']".


vrana:
Merged. I've changed the index type from PARTIAL to INDEX.

PR #207, opened 2017-02-19
Fixed all alternative designs after changes in a list of tables

peterpp:
Fixed all alternative designs after changes in a list of tables made here: https://github.com/vrana/adminer/pull/116


vrana:
This must be lots of works, thanks.

PR #209, opened 2017-02-23
Login screen: recognize mysql URLs

jmfayard:
In the login screen, entering as server an URL like

  mysql://user:pass@example.com/my_db?options [1]

is a shorthand to setting 

- username=user, 
- password=pass,
- server=example.com, 
- database=my_db

[1] this syntax is used for example by doctrine


vrana:
Please implement this as a plugin.

PR #212, opened 2017-04-06
Fix PostgreSQL nullable fields in export

Deni42:
When exporting the database schema, nullable fields show as non nullable and non nullables show as nullable. This commit fixes the issue.


vrana:
Merged, thanks.

kusmierz:
@vrana, @Deni42 - because of 8647f1aeb32d7d24c4e2488660677ce4d7a9c9be commit, the list of tables (aka "menu") is ordered by old instead of table's names...

Deni42:
I changed the table ordering because when you use foreign keys, they need to be imported in the correct order, you can't create a table that references a table that doesn't exist. A perfect solution would look up all dependencies and create the correct order, but for now, it was just easier to export by creation date (= object id).

I guess the table selection menu should use a different query, or use a parameter for the function. Feel free to improve it.

vrana:
Reverted.

PR #215, opened 2017-05-03
Update AdminerTablesFilter to produce li as per c9801e2e2418

jnvsor:


jnvsor:
Noticed this was wrong when I updated Haeckel and lost icons

vrana:
Merged, thanks.

PR #216, opened 2017-05-25
Update adminerTablesFilter to set up filter asap

jnvsor:
If you're opening a very large database on a very slow server,
adminer might take a few seconds to load the whole table list.

Because the JS for the filter is added after the tables are
shown this can result in the user typing into the filter and
nothing happening, followed by the filter contents being replaced
by the sessionStorage cached contents.

This is confusing and annoying, so this commit does most of the
work beforehand and only leaves triggering the filtering itself
until after the whole list of tables has been printed.

### DIFF
--- a/plugins/tables-filter.php
+++ b/plugins/tables-filter.php
@@ -9,20 +9,6 @@
 class AdminerTablesFilter {
 	function tablesPrint($tables) { ?>
 <p class="jsonly"><input id="filter-field" onkeyup="tablesFilterInput();" autocomplete="off">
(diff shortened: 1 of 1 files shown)


djmattyg007:
This has been bothering me, I'd love to see this merged.

jnvsor:
Perhaps we should address the underlying question of "What happened in 4.3 that made listing the tables so much slower" but the bug tracker on SF is unusable :)

djmattyg007:
I do wish the issue tracker was open here. It's always seemed odd that sourceforge was still in use.

vrana:
Should be fixed by 2dcad1f2.

PR #219, opened 2017-07-24
Skin: Chrome issue fix

jiripavlicek:
Hi Jakub, there is a fix of Chrome related issue - shifted icon for "select" and table name in left tables list. In Firefox it's fine.


vrana:
I see no icon there. Neither in Firefox nor in Chrome.

PR #221, opened 2017-08-15
Add text input to change filename

mikaelz:
On shared hostings the database names are autogenerated. It would be great to change dump filename when setting export options or be able to add some prefix/suffix.

ss_2017-08-15-11-57-47


mikaelz:
Hello Jakub @vrana, could you please review?

vrana:
I am not interested in having this in Adminer. It could probably be implemented as a plugin though. Please send a pull request if not.

PR #222, opened 2017-08-23
Change animated icons/links to static icons

jkufner:
The animated links SQL/import/export/create table are extremely annoying. This changes them to nice static buttons. Not too big, not too small.

screenshot_20170824_013350


vrana:
I guess this is there by purpose.

PR #225, opened 2017-09-17
add schema search path for query as an extendable option

jlr2k8:
As posted here, https://sourceforge.net/p/adminer/discussion/960419/thread/e7a8672e/, adding multiple schemas to a search path (PostgreSQL) would allow shorter queries to be written in the SQL command textarea. Currently, only one search path can be utilized at a time. This new functionality allows comma separated schemas to be entered when writing a query, and is available by extending within the Adminer object.

### DIFF
--- a/adminer/drivers/pgsql.inc.php
+++ b/adminer/drivers/pgsql.inc.php
@@ -287,10 +287,12 @@ ORDER BY 1";
 	}
 
 	function table_status($name = "") {
+        global $connection;
+        $schema = !empty($_GET['ns']) ? "'" . pg_escape_string($connection->_link, $_GET['ns']) . "'" : 'current_schema()';
 		$return = array();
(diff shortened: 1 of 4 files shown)


vrana:
The `SET search_path TO ` can be called directly in the SQL command. I'm not going to accept this pull request.

jlr2k8:
Right, that is true. This simply assists with a specific use case at our workplace, and I thought it'd be nice to share.

This functionality simply allows the search path to be set ahead of the query so it doesn't have to be specified within the query itself. PhpPgAdmin does exactly that, so this is really just a way to mimic their functionality.

At my workplace, we currently use Adminer with this added functionality (pull request approved or not). We have dozens of schemas, each filled with many tables - so it is nice to be able to write a workable query (which can join tables in various schemas) without having to specify the schema for every table in that query.

seidlmic:
I would welcome this functionality too.

cvvkila:
I have a similar need for Postgresql, but not in the editor. On the left panel, you can select schemas (information_schema, public, postgres...) and push export  button and all tables within this schema will be exported. But some users would like to export all tables for all there schemas without selecting one by one.  That is to say having a joker character entry to export the schemas they are granted to access. 
I don't know if this is achievable with a plugin?

vrana:
Adminer 5.4.4 supports exporting all or selected schemas in the database at once.

PR #229, opened 2017-09-27
Add ability to override file name (path) when loading sql dump from webserver

noandrea:
Allows to set the name/path for loading sql dump from webserver local file in adminer configuration


vrana:
What's the rationale behind this? Why would a customization like to change it?

noandrea:
we use docker to have a shared standard environment for web projects. 
each docker image comes with adminer as the default database interface, and each docker setup has a standard sql file that is versioned and shipped with the project . 

a typical case is when we need to publish changes on the staging server, the sql file is already on the server and since there is already the button "run file from webserver" it will be very convenient to have it configured to load the right sql

vrana:
I've renamed the method to `importServerPath` and merged this.

PR #231, opened 2017-10-18
Enhance ElasticSearch integration

fabiomontefuscolo:
Hello!

I'm using Adminer to take care of my data indexed on ElasticSearch. There are some methods sometimes I need, like delete, edit and insert records, but is actually missing.

This pull request is a proposal for those missing methods and a small bug fix to show counts for each **type**. Could you check it?

Thank you in advance!
Cheers


fabiomontefuscolo:
There are some problems with your travis environment. My changes was accepted for php 7.1 but is broken for 5.3. The message is:

```
0.02s$ phpenv global 5.3
rbenv: version `5.3' not installed
The command "phpenv global 5.3" failed and exited with 1 during .
Your build has been stopped.
```

marclaporte:
For the record, this is how we use it: Adminer is shipped as part of the Elasticsearch app for ClearOS:
http://wikisuite.org/How-to-install-Elasticsearch-on-ClearOS

Thank you Adminer!

adminer/drivers/elastic.inc.php
```diff
@@ -292,15 +301,22 @@ function tables_list() {
 
 	function table_status($name = "", $fast = false) {
 		global $connection;
+<<<<<<< HEAD
 		$search = $connection->query("_search?search_type=count", array(
+=======
+
+		$search = $connection->query("_search", array(
+			"size" => 0,
+>>>>>>> elastic
```
djmattyg007: You've committed a merge conflict.

fabiomontefuscolo: Sorry! I think I messed .travis.yml also, so I reverted it, but adding two commands on script section to check also the compiled file. Because on first of my commits, the code passed on linter, but compiler could not generate a good compiled file.

vrana:
Thanks for your work. I don't use ElasticSearch so I've merged this without testing. I've fixed the style to be same as in the rest of Adminer. Please create a separate PR for the Travis change. It seems to me that the result of `test $? -eq 1` would be ignored but I'm not an expert in Travis.

PR #232, opened 2017-10-22
Allow importing bzip2 compressed files

jtojnar:


vrana:
There's also import in `sql.inc.php`, please modify it as well.

jtojnar:
Would this work? I do not really like the serial check (nested ternaries) but there is not much room for improvement.

vrana:
Bzip2 is not supported in export. There's a plugin for Bzip2 export and I probably don't want to integrate it into Adminer. To avoid asymmetry, I am not going to integrate this either. Sorry about asking you to do an extra work.

PR #233, opened 2017-11-02
Enforce sqlite foreign keys

mcorne:
Sqlite does not enforce foreign key constrains by default, it has to be done every time the database is open.
Note that the "PRAGMA foreign_keys = ON" must be executed in the constructor instead of the query() method. Otherwise it could conflict with a multi query call (import, SQL query) containing transactions or foreign keys off pragma such as a database creation script. This pragma must be executed outside of any transaction according the Sqlite doc.


vrana:
SQLite can be used via PDO, this patch doesn't address it. I'll add the support but not by this patch.

PR #234, opened 2017-11-02
Prevent display of null option for non null foreign keys

mcorne:
The null (empty) option is proposed for non null foreign keys which, on insert or update, generates a foreign key constraint error or an inconsistency if the constraint is disabled. The null option should only be displayed for foreign keys that may be null.


vrana:
It's better to get an error if I forget to select something then to get the first value.

PR #235, opened 2017-11-24
Hash column

mcorne:
This plugin hashes a column ending with "_md5", "_sha1", "_hash" etc. which is useful for passwords.

### DIFF
--- a/plugins/hash-column.php
+++ b/plugins/hash-column.php
@@ -1,0 +1,192 @@
+<?php
+/** Hash fields ending with an algo name suffix
+ *
+ * Hash fields ending with "_md5", "_sha1", "sha512_224", "tiger128_3" etc. with the hash() function.
+ * Hash fields ending "_crypt" with the crypt() function (for PHP < 5.5).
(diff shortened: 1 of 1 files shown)


plugins/hash-column.php
```diff
@@ -0,0 +1,192 @@
+<?php
+/** Hash fields ending with an algo name suffix
+ *
+ * Hash fields ending with "_md5", "_sha1", "sha512_224", "tiger128_3" etc. with the hash() function.
+ * Hash fields ending "_crypt" with the crypt() function (for PHP < 5.5).
+ * Hash fields ending with "_hash" with the password_hash() function (recommended for PHP 5.5 and above).
+ *
+ * Enter "<password>[::<salt>]" when editing a column to be hashed with crypt(), ex. "some text::some salt".
+ * The salt is optional although strongly recommended.
```
vrana: Please remove this part and auto-generate the salt as a random string which is recommended for storing passwords. It would also simplify the usage.

PR #236, opened 2017-12-11
added mongodb PHP7 beta driver.

morozovsk:
Work only insert, update, list database, list index, list tables. select, where, limit, order, delete, edit, edit from select.
does not work structure modify, index, modify, custom queries, export, import, groups, collections drop, add.

### DIFF
--- a/adminer/drivers/mongo.inc.php
+++ b/adminer/drivers/mongo.inc.php
@@ -147,29 +147,6 @@ if (isset($_GET["mongo"])) {
 		}
 	}
 
-
-
-	function connect() {
-		global $adminer;
-		$connection = new Min_DB;
-		$credentials = $adminer->credentials();
-		if ($connection->connect($credentials[0], $credentials[1], $credentials[2])) {
-			return $connection;
-		}
-		return $connection->error;
-	}
-
-	function error() {
-		global $connection;
-		return h($connection->error);
-	}
-
-	function logged_user() {
-		global $adminer;
-		$credentials = $adminer->credentials();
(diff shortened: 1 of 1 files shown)


morozovsk:
it has nothing in common: https://gist.github.com/morozovsk/14374da65d88f32f94ad061f4a7b942c/revisions

adminer/drivers/mongo.inc.php
```diff
@@ -354,4 +353,614 @@ function support($feature) {
...
+						}
+
+					}
+				}
+			}
+
+			return $data;
+		}
+
+		$jush = "mongodb";
```
vrana: Use `mongo`.

morozovsk:
>Most of the global functions (connect(), error(), …, support()) and variables ($functions, …, $edit_functions) seem to be copy/pasted.

I see you prefer 100 "if/else" than one "if/else".
>Use mongo.

-> use php5
>mongo is legacy now: 

http://php.net/manual/en/book.mongo.php
>This extension is deprecated. Instead, **the MongoDB extension should be used**.

http://php.net/manual/en/mongo.installation.php

vrana:
Put all common functions in a common block. Put extension-specific functions in the `if` block.

JUSH is a syntax highlighter, `$jush` holds the name of the language. It has nothing in common with PHP version or extensions.

morozovsk:
done

vrana:
Merged, thanks.

morozovsk:
Thank you @pivolan for your help!

PR #167, opened 2016-01-07
Bosnian language

rapttor:
My contribution for this great piece of code. Bosnian translation for ADMINER

PR #237, opened 2017-12-18
MSSQL driver fixed for freetds drivers. See bug #167, #291

uhi67:
Sometimes (for example on linux servers) there are no other MSSQL driver than freetds. In this case, the query of the table structure may fail.

In the function fields(), line 364 the sql command `"SELECT c.*, t.name type, d.definition [default] FROM ... ` is not good for freetds driver.
If you change it to
```
    "SELECT 
        c.max_length, c.precision, c.scale, c.name, c.is_nullable,
        c.is_identity, c.collation_name, 
        t.name type, CAST(d.definition as text) [default]"
```
the UNICODE error will disapper, and table structure will be shown appropriately.

### DIFF
--- a/adminer/drivers/mssql.inc.php
+++ b/adminer/drivers/mssql.inc.php
@@ -361,7 +361,9 @@ if (isset($_GET["mssql"])) {
 
 	function fields($table) {
 		$return = array();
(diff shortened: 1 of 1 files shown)


vrana:
Merged, thanks.

PR #238, opened 2017-12-22
Add support to cascate on drop schema with new button

webs86:
I have added a new button to alter schema page to support DROP SCHEMA CASCADE on postgres


vrana:
This seems a little bit dangerous.

PR #239, opened 2018-01-02
feat(adminer/static/function.js): add menuAutoScroll to improve the e…

TIGERB:
Using the adminer more than two years, which is a very cool product. but when a database's table are too much, the tables menu can't be scrolled, and the selected table can't scroll to the current position when refresh the page. The efficiency in the practical development experience is not good, so I submit the PR.

### DIFF
--- a/adminer/include/adminer.inc.php
+++ b/adminer/include/adminer.inc.php
@@ -951,13 +951,13 @@ bodyLoad('<?php echo (is_object($connection) ? substr($connection->server_info, 
 	* @return null
 	*/
 	function tablesPrint($tables) {
-		echo "<ul id='tables' onmouseover='menuOver(this, event);' onmouseout='menuOut(this);'>\n";
+		echo "<ul id='tables' onmouseover='menuOver(this, event);' onmouseout='menuOut(this);' style='height: 600px;'>\n";
 		foreach ($tables as $table => $status) {
 			echo '<li><a href="' . h(ME) . 'select=' . urlencode($table) . '"' . bold($_GET["select"] == $table || $_GET["edit"] == $table, "select") . ">" . lang('select') . "</a> ";
 			$name = $this->tableName($status);
 			echo (support("table") || support("indexes")
 				? '<a href="' . h(ME) . 'table=' . urlencode($table) . '"'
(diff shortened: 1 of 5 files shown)


adminer/include/adminer.inc.php
```diff
@@ -951,13 +951,13 @@ function databasesPrint($missing) {
 	* @return null
 	*/
 	function tablesPrint($tables) {
-		echo "<ul id='tables' onmouseover='menuOver(this, event);' onmouseout='menuOut(this);'>\n";
+		echo "<ul id='tables' onmouseover='menuOver(this, event);' onmouseout='menuOut(this);' style='height: 600px;'>\n";
```
vrana: I don't like the fixed height. What if the display has another height?

adminer/include/adminer.inc.php
```diff
@@ -951,13 +951,13 @@ function databasesPrint($missing) {
...
 		foreach ($tables as $table => $status) {
 			echo '<li><a href="' . h(ME) . 'select=' . urlencode($table) . '"' . bold($_GET["select"] == $table || $_GET["edit"] == $table, "select") . ">" . lang('select') . "</a> ";
 			$name = $this->tableName($status);
 			echo (support("table") || support("indexes")
 				? '<a href="' . h(ME) . 'table=' . urlencode($table) . '"'
-					. bold(in_array($table, array($_GET["table"], $_GET["create"], $_GET["indexes"], $_GET["foreign"], $_GET["trigger"])), (is_view($status) ? "view" : "structure"))
+					. bold(in_array($table, array($_GET["table"], $_GET["create"], $_GET["indexes"], $_GET["foreign"], $_GET["trigger"])), (is_view($status) ? "view" : "structure $name"))
```
vrana: What's this, how it is related?

adminer/static/functions.js
```diff
@@ -280,18 +280,37 @@ function pageClick(href, page, event) {
...
+		el.style.overflow = 'scroll';
 	}
 }
 
 /** Hide items in menu
 * @param HTMLElement
 */
 function menuOut(el) {
-	el.style.overflow = 'auto';
+	el.style.overflow = 'scroll';
```
vrana: Why do you change it to `scroll` here and also to `scroll` in `menuOver`?

vrana:
I am not interested in this but it could probably be implemented as a plugin.

PR #241, opened 2018-01-19
adminer-theme-mancave-0.1

monolithforge:
Thanks Adminer!!
    
    I've used this mancave theme for 10+ years that.
    It's a dark theme, but it's always my goto theme- which is why I feel I should share it with you all.
    I hope others get the same satysfaction out of it that we have (-:

### DIFF
--- a/designs/mancave/adminer.css
+++ b/designs/mancave/adminer.css
@@ -1,0 +1,231 @@
+/*
+  
+  VERSION: adminer-theme-mancave-0.1
+  
+  AUTHORS: panreach@gmail.com, dev@monolithforge.com //NOTE CREDITS BELOW... that we took from ;-)
+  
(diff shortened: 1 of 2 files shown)


monolithforge:
FYI @vrana 

P.S.  Could you create an upstream dev branch of adminer-themes on master?  I have hack'd your script(s) to include a themes dropdown via: https://github.com/panreach/adminer/releases/tag/adminer-themes-alpha-4.3.2-dev

It's not nearly close to being minimized (and it is very large for a single file), but it (maybe) might have promise for offering (others) themes in one file?

-panreach

P.S.  Don't want to step on any toes, but I'm a long time user of Adminer over PHPMyAdmin.  Thanks for making my day job so much easier!!

vrana:
Thanks, I've merged mancave2.

PR #242, opened 2018-01-20
 Unify HTML for list of tables in Editor according to Adminer 

peterpp:
Adminer uses `<ul>` for list of tables, so Editor should too (for compatibility with CSS).


peterpp:
I will create new PR with just one commit. Sorry for the mess.

vrana:
Merged, thanks.

PR #243, opened 2018-01-21
Encapsulate fulltext part in search box into <div>

peterpp:
Search box contains a <div> for each condition. However, fulltext condition is printed inline with <br> at the end. This causes problems when I want to set CSS for search box parts.


vrana:
Merged, thanks.

PR #246, opened 2018-01-26
fix elastic.inc.php to work on PHP 5.2

helix84:
Without this change, I get the following syntax error on PHP 5.2.6:
PHP Parse error:  syntax error, unexpected ':' in /var/www/adminer/adminer/drivers/elastic.inc.php on line 415

I haven't actually tested the alter_table() method.


vrana:
Merged, thanks.

PR #251, opened 2018-02-08
version_compare only use string.

soleuu:
Hello,

this pull request fix the following error (regression since 4.6.0) in processList:

ERROR: column "procpid" does not exist
LINE 1: SELECT * FROM pg_stat_activity ORDER BY procpid

Explanation : version_compare use string and not real nor integer.
the automatic cast made at call time seems to cast 9.2 in '9.199999999999...'
We use postgresql 9.5.10 so  19999... is bigger tant 2. 

I have made the change only for portgresql.


vrana:
Fixed by setting `precision` to 15 instead. This "rounding" would cause troubles also when returning 9.2 from SQLite.

PR #252, opened 2018-02-09
Fixing exports on PostgreSQL

PotatoPangolin:
Since PostgreSQL 10, sequence info have been moved to a new catalog **PG_SEQUENCES** instead of being in the sequence itself, so the export was broken.

Also, not related to the version 10 of PostgreSQL but the export for **character varying** fields didn't put quotes around the default values.

### DIFF
--- a/adminer/drivers/pgsql.inc.php
+++ b/adminer/drivers/pgsql.inc.php
@@ -704,7 +704,7 @@ AND typelem = 0"
 		// fields' definitions
 		foreach ($fields as $field_name => $field) {
 			$part = idf_escape($field['field']) . ' ' . $field['full_type']
-				. (is_null($field['default']) ? "" : " DEFAULT $field[default]")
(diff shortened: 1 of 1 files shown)


vrana:
For reference, the change is documented at https://www.postgresql.org/docs/10/static/release-10.html#id-1.11.6.7.5.

PotatoPangolin:
Thanks. According to this link, in E.3.2. on the 9th bullet :
> To obtain the sequence's other properties, applications must look into `pg_sequence`
So I guess `pg_sequence` should be used instead of `pg_sequences` or `information_schema.sequences`.

I'll move the unrelated part.

vrana:
I've fixed this differently, please verify that it works as I don't have PostgreSQL 10.

vrana:
I've fixed the default value issue too. Thanks for reporting.

PR #255, opened 2018-02-16
Aggregate and time functions added to PgSql driver

sartor:
For PostgreSQL driver added:
Functions:
1) extract (epoch from $column)
2) date

Aggregation functions:
1) ARRAY_AGG
2) JSONB_AGG
3) STRING_AGG
4) STRING_AGG (DISTINCT)


vrana:
I'm not interested in adding every possible function. Only those that are frequently used.

PR #256, opened 2018-02-22
Update adminer-design (rmsoft) to version 4.6.2.1

mesaros:
Update adminer-design (rmsoft) to version 4.6.2.1


vrana:
Merged, thanks. I've wrapped the lines so that this is diffable.

PR #257, opened 2018-02-25
Add missing variables to global declaration

perob:
These two variables: $driver and $operators, seems to be missed from global variables declaration.

I needed $driver when working with Sqlite database and $operators seems to be used with MongoDB.

As from source file comment:
"allows including Adminer inside a function"

### DIFF
--- a/adminer/include/bootstrap.inc.php
+++ b/adminer/include/bootstrap.inc.php
@@ -33,7 +33,7 @@ if ($_GET["script"] == "version") {
 	exit;
 }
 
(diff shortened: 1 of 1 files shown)


vrana:
I've fixed MongoDB to use `$adminer->operators` and merged the `$driver` change.

brusch:
@vrana when is the next patch release going to be released, is there already a date? Would be quite good, since this bug breaks Adminer in our case. Thank you so much!

PR #258, opened 2018-02-27
Mongodb: implement create_database

David-5-1:
Create database fails on MongoDB just because the method is not found although there is no need to actually do anything as MongoDB creates a database only when actual data is written inside it.

### DIFF
--- a/adminer/drivers/mongo.inc.php
+++ b/adminer/drivers/mongo.inc.php
@@ -28,7 +28,7 @@ if (isset($_GET["mongo"])) {
 					return false;
 				}
 			}
-			
+
 			function query($query) {
 				return false;
 			}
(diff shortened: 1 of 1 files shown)


vrana:
Please avoid commits to remove tabs on empty lines, Adminer coding style allows it.

Instead of doing nothing and reporting success, the links to create or rename the database should not even show. You can achieve it by adding a new entry to `support()`, supporting it in all drivers except MongoDB and wrapping the code to check it first.

vrana:
This actually makes sense as MongoDB lists existing databases. Merged.

David-5-1:
Thanks for the reply.

Since what I did do not really creates a DB, the DB name does not appear in left-side bar until a collection is actually created. I am not sure if/how it can be fixed.

vrana:
I've fixed it by displaying the current database in the list even if it doesn't exist. It's useful for other drivers too.

PR #259, opened 2018-02-28
Do not cast date/timestamp exact searches in PostgreSQL to text

finwe:
This fixes faulty behaviour when comparing date/timestamp fields with `=`, when value casted to text did not match simple 'YYYY-MM-DD date notation'.


vrana:
Merged and added `time` too.

PR #260, opened 2018-03-01
When dev and stable version are same but update is shown - fixed

thamaraiselvam:
unit tested with following test cases

Current version 4.6.2-dev
Stable version  4.6.2 

Current version 4.6.1-dev
Stable version  4.6.2 

Current version 4.6.2-dev
Stable version  4.6.1


vrana:
4.6.2-dev < 4.6.2 so you should update.

PR #261, opened 2018-03-02
support VACUUM FULL for pgsql

pipcet:
This minor change adds a "Full Vacuum" button which executes VACUUM FULL on a PostgreSQL table, thus reclaiming extra disk space it might be using.


vrana:
I'm not interested.

PR #262, opened 2018-03-07
PgSQL convert search fix

sartor:
PgSQL search cast fix.
Problem was when searching over integer columns: wrong cast to text was added
Looks like invert typo occurred


vrana:
If `val` is numeric then `int` is added. I don't see what's the purpose of this change. Does it fix a specific problem? Which one?

sartor:
When in search box select operator IN for any integer field like:
id IN (1,2,3)
Is converts to:
CAST (id as 'text') IN (1,2,3)
This fix changes this to valid (without casting):
id IN (1,2,3)

vrana:
I see. I just added the `number_type()` unconditionally. I also improved the _anywhere_ search to support comma separated numbers with `IN`.

sartor:
Thank you!

PR #263, opened 2018-03-07
PgSQL search operator "SQL" added

sartor:
SQL operator allowed for PostgreSQL.
Warning: Possibility to drop table in select query!
But since adminer is a fairly low-level application, this option will be more useful than dangerous, imho


vrana:
Merged, thanks.

vrana:
I have to revert this change, it is vulnerable to CSRF. We would need to change the search to send the CSRF token.

PR #265, opened 2018-03-20
set global $driver

Gerhard-Kanzler:
becaus there is an error: tableHelp on null (mysql)


vrana:
Already solved by 6c63c2f9.

Gerhard-Kanzler:
The Problem is not solved sorry!

You need "global $driver" Please add this line, In Pimcore when you use the adminer and open a table you get this error.

vrana:
acee5c00 wasn't released yet.

PR #267, opened 2018-03-23
Fix two errors for PHP 7.2

marcdejonge:
It seems that PHP 7.2 is a bit more strict. For example, I couldn't perform a normal select, because the `$driver` global variable was `null`. When adding the global definition when creating it, it will be solved. Also, you need to call `more_results()` before calling `next_result()`. This is a warning that in our environment  that resulted in an error.

I've included adminer in our own tool, that's why we see these warnings.

### DIFF
--- a/adminer/include/auth.inc.php
+++ b/adminer/include/auth.inc.php
@@ -167,7 +167,6 @@ if (isset($_GET["username"])) {
 	}
 	check_invalid_login();
 	$connection = connect();
-	global $driver;
 	$driver = new Min_Driver($connection);
 }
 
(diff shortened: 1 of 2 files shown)


vrana:
`$driver` is already marked as global since acee5c00. Revert this part.

Which extension requires calling `more_results` before `next_result`?

marcdejonge:
I got a warning of the normal MySQL driver that the more_results needed to be checked before calling next_result.

vrana:
According to http://php.net/mysqli-stmt.next-result#refsect1-mysqli-stmt.next-result-errors, this emits only `E_STRICT` and Adminer ignores these.

PR #268, opened 2018-03-27
MSSQL: Fixes (connection, columns)

blackiceua:
Added fix for connection through sqlsrv driver (added possibility to use port)
https://docs.microsoft.com/en-us/sql/connect/php/how-to-connect-on-a-specified-port

Added fix for SSMA_TimeStamp column (it's updated automatically with every change).

Added additional params to the function dumpTable in the Editor.

### DIFF
--- a/adminer/drivers/mssql.inc.php
+++ b/adminer/drivers/mssql.inc.php
@@ -24,7 +24,14 @@ if (isset($_GET["mssql"])) {
 			}
 
 			function connect($server, $username, $password) {
-				$this->_link = @sqlsrv_connect($server, array("UID" => $username, "PWD" => $password, "CharacterSet" => "UTF-8"));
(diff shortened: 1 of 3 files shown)


vrana:
I've included the port support (simplified).

its0x08:
Thats exactly what Im missing right now. The port feature for MySQL, since my dev envoirment uses Docker they deployed multiple dbs for testing in myltiple ports therefore I cannot connect to any of them except for the 3306!!

vrana:
> Thats exactly what Im missing right now. The port feature for MySQL, since my dev envoirment uses Docker they deployed multiple dbs for testing in myltiple ports therefore I cannot connect to any of them except for the 3306!!

You can specify the port in the server field using `server.name:3307`.

PR #269, opened 2018-04-05
[View] [Export] Fix export problem when having views

dinbrca:
When having views - the export table would break like this:
screen shot 2018-04-05 at 9 40 03 am

Now it's this:
screen shot 2018-04-05 at 9 39 40 am


vrana:
This doesn't follow the minimalist nature of Adminer. Feel free to update the skin to show the border.

PR #270, opened 2018-04-08
Respect 'session.cookie_secure' if it is On

FranklinWhale:
If session.cookie_secure is On already, it should not be overridden to Off because the HTTPS server variable is not set.


vrana:
The `$HTTPS` variable stores the information whether the page was server over HTTPS. `session.cookie_secure` doesn't provide this information. If the page was not served over HTTPS then requiring the cookie to be secure would not work.

FranklinWhale:
The issue is that the php server may be behind a proxy server and not handling HTTPS traffic directly. In this case, the https variable is not set. Since `session.cookie_secure` is off by default, when it is on, it must be explicitly set by a human. I think such preference should be preserved.

vrana:
Good point, merged with an added comment.

PR #271, opened 2018-04-20
Prepopulate list of known SQLite databases

mikhailmelnik:
Even with SQLite plugin installed Adminer does not provide the same functionality as for the Postgres such as listing of accessible databases. Although it is not possible with SQLite it will be handy to specify list of databases you work often with.


mikhailmelnik:
Went through the codebase again and even if this thing is handy it does not belong here, sorry.

PR #272, opened 2018-05-02
Allow AdminerTablesFilter plugin to be used with third-party table list plugins

peterpp:
AdminerTablesFilter now doesn't render custom list of tables. This will allow to create plugin for rendering list of table compatible with filter plugin.


vrana:
I'm not interested in adding `data` attributes just for plugins.

peterpp:
OK, if you think that reimplementing whole list of tables in plugin is better. I don't.

PR #273, opened 2018-05-05
Elasticsearch Enhancements

fabiomontefuscolo:
Hi,

I'm working with @marclaporte to enhance integration between adminer and elasticsearch. This pull request intends to:

* fix search on elasticsearch >= 5.x
* fix search from tables list page

In some cases, SQL queries are passed to `$connection` object and I had no chance to change this behavior without change other parts of adminer. So I decided to create a small filter system to intercept and change parameters sent to query. On this way, I could detach some ""hacky"" code from `$connection`.

Please, take a look on this pull request and tell me what is needed to be accepted.

Thank you very much for adminer! :)

### DIFF
--- a/adminer/drivers/elastic.inc.php
+++ b/adminer/drivers/elastic.inc.php
@@ -5,6 +5,62 @@ if (isset($_GET["elastic"])) {
 	$possible_drivers = array("json");
 	define("DRIVER", "elastic");
 
+	class AdminerDriverFilter {
+		private static $instance = null;
+		private $filters = null;
+
+		private function __construct() {
+			$this->filters = array();
+		}
+
+		public static function getInstance() {
+			if (self::$instance === null) {
+				self::$instance = new self();
+			}
+			return self::$instance;
+		}
+
+		public function add($name, $callback, $priority=10) {
+			if (empty($this->filters[ $name ])) {
+				$this->filters[ $name ] = array();
+			}
+			$this->filters[ $name ][ ] = array($callback, $priority);
+		}
+
+		public function run($name, $subject) {
+			if (empty($this->filters[ $name ])) {
+				return $subject;
+			}
+			foreach ($this->filters[ $name ] as $filter) {
+				list($callback, $priority) = $filter;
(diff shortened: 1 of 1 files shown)


fabiomontefuscolo:
Also, I think it would be nice to have that Filter[1] class available on entire adminer, so we could extend adminer behavior more easily.

[1] https://github.com/tikiorg/adminer/blob/master/adminer/drivers/elastic.inc.php#L13

vrana:
I don't like the filter. Adminer has enough abstractions already, do not add a new one.

Follow coding style used in the rest of the code (no spaces inside `()` or `[]`, no variable names in `@param`, no double spaces anywhere, space in `if (`, … Be consistent.

Do not introduce another global variable.

Create pull requests for individual fixes.

fabiomontefuscolo:
Sure, I will fix it and create another pull request. But can you help at one point? I created the filter to dettach a hack from connection driver and make possible to easily undo it, but without that, I don't know how to fix the search on Table List page.

The exactly point of my doubt is at `function.php` line 87. The `table_status` send a SQL directly on `$connection->query`. I did this code to ""guess"" user is searching on all tables and rewrite the query https://github.com/tikiorg/adminer/blob/master/adminer/drivers/elastic.inc.php#L631.

What do you suggest? Any clue??

image

Thank you for sharing your time!!

marclaporte:
@vrana: we want to improve, but we are stuck and need your guidance.

Thanks!

vrana:
There's no `function.php` and there's nothing relevant at line 87 in `include/functions.inc.php`.

`search_tables` calls `$connection->query`. Instead it could call `$driver->searchTables` with the default implementation in `Min_SQL` calling the current `$connection->query`. You would then override this in `elastic.inc.php`.

You will need to touch these files:
`functions.inc.php`: Change `$connection->query` to `$driver->searchTables`.
`driver.inc.php`: Add `searchTables` calling the original `$connection->query`.
`elastic.inc.php`: Add `searchTables` with anything you need.

PR #274, opened 2018-05-07
on Ubuntu 18.04 and MariaDB 10.3.6 database list is unsroter

bbaronSVK:


vrana:
Should be fixed by 11f24a52.

PR #277, opened 2018-05-08
Add class 'data' to all non-layout tables

peterpp:
Tables that display data can have different design than tables used just for layout.


vrana:
Which tables are for layout? Can you add a class to them instead?

peterpp:
Layout tables are used in forms (for example login form). We discussed it in email and you confirmed that I can add class "data" to all tables with <thead>. However, I can add class to layout tables as well.

vrana:
There are 33 `<table>` and 20 `<thead>` so adding it to tables without `<thead>` should be less work.

PR #278, opened 2018-05-09
Fix table list in sidebar in designs/pokorny

piit79:


piit79:
There is a bug in the table list in Chrome in the pokorny design - the "select" icons are not vertically aligned with the table name as shown below on the left. (The issue is not present if Firefox.) This commit fixes the issue as shown below on the right. Fix also tested in Firefox.
pokorny-chrome-bug pokorny-chrome-fix

PR #280, opened 2018-05-17
Fixes foreign key pattern in MySQL

jrdbrndt:
Foreign keys are not displaying in MySQL when ANSI_QUOTES are enabled because the preg_match_all() is looking specifically for backticks. This fix allows for double quotes as well, so foreign keys are recognized with ANSI_QUOTES enabled.


jrdbrndt:
I shouldn't be committing to master. My apologies. I will branch and resend the request.

PR #281, opened 2018-05-17
Fixes foreign key pattern in MySQL

jrdbrndt:
Foreign keys are not displaying in MySQL when ANSI_QUOTES is enabled because the preg_match_all() is looking specifically for backticks. This fix allows for double quotes as well, so foreign keys are recognized even when ANSI_QUOTES is enabled.


vrana:
Merged, thanks.

PR #284, opened 2018-06-25
prevent "cannot change cache limiter when session is active" warning

staabm:
use a more rebust "session is alive" check to prevent "cannot change cache limiter when session is active" warnings (Bug 625)

https://sourceforge.net/p/adminer/bugs-and-features/625/


staabm:
just for your information. it seems the latest adminer version (4.6.3) does not have the issue I reported initially.

In case you feel this PR makes the code nevertheless more readable, feel free to merge it.
if not, I am also ok to close it and leave everything as is.

vrana:
`session_status` is PHP >= 5.4.0, Adminer supports earlier versions. Also, I don't see how `SID` could be undefined here.

PR #285, opened 2018-06-26
Add Without data checkbox in sql command ui

adhocore:
This helps in quickly testing how long queries take and error (if any) without dumping tons of data in the ui, especially for select queries

(We can use already available `Limit rows: 1` but still there's some data, while this is just handy)


vrana:
There would be too many checkboxes. Maybe “Show only errors” should control this? After all, `SELECT` output is not an error.

adhocore:
alright

PR #287, opened 2018-07-15
Adminer theme

alekseymvt:
New adminer theme.
https://github.com/alekseymvt/Adminer.theme/raw/master/adm.png


vrana:
Thanks, added to https://www.adminer.org/#extras.

PR #289, opened 2018-07-17
DOM not ready when restoring `adminer_tables_filter`

Fishdrowned:
Uncaught TypeError: Cannot set property 'value' of null


Fishdrowned:
Anybody?

alekseymvt:
Probably on vacation :)

PR #293, opened 2018-07-27
Add class 'layout' to all non-data tables

peterpp:
This class will allow to change design of tables that are used just for layout inside forms.


vrana:
Merged.

PR #294, opened 2018-07-27
Links

alekseymvt:
* Add class "link_xxx" to "[p class='links']" child elements.
* Refactoring function bold.


vrana:
Please avoid random unrelated changes (`</p>`, changing return type, lines consisting of whitespace). I am not interested in the `bold` function change, I might consider the `link_xxx` if you split it in a separate pull request.

PR #298, opened 2018-08-27
Adminer mode

alekseymvt:
Global constant, informing in what mode the adminer works  (adminer/editor).


vrana:
This shouldn't be needed. Editor is just a customization of Adminer (and even Adminer is its own customization). There could be any other customization.

In other words, this is not general enough.

PR #299, opened 2018-09-01
Update zh-tw.inc.php

onecentlin:
Fix some typo and add white spaces for readability


vrana:
Please get one vote from any user that this is better than the original.

onecentlin:
@vrana Got votes with thumbs up.

PR #300, opened 2018-09-05
add search textbox for tables

riomunas:
Add search textbox for tables;
when keyup on textbox it will show only table with maching criteria


vrana:
There's already a plugin for this.

PR #303, opened 2018-09-21
Hide import from server if no path

ilessiivi:
Adds the ability to hide the import from server functionality, for installs where Adminer runs in a location where the user can't upload files.


vrana:
Merged, thanks.

PR #304, opened 2018-10-04
call adminer from plugin

Artistan:
allow plugins to reference the adminer instance so that any other registered plugin will still be called.. 

Example...

this plugin calls loginForm on the adminer instance after it is called itself.  
this allows for other registered plugins with adminer to work within that call (loginFormField)  

```php
<?php
class AdminerFormWrapper
{
	var $adminer;
	public function __construct(&$adminer)
	{
		$this->adminer =& $adminer;
	}
	function loginForm()
	{
                // prevent plugins from being recursively called...
                $this->adminer->propagation=false;
		?>
		<div id="login-form">
			<?php $this->adminer->loginForm() ?>
		</div>
		<?php
		return true;
	}
}

function adminer_object()
{
	// Required to run any plugin
	include_once './plugins/plugin.php';
	// Plugins auto-loader
	foreach (glob('plugins/*.php') as $filename)
	{
		include_once "./{$filename}";
	}
	// Specify enabled plugins here
	$plugins = [
/** plugins that may have loginFormField defined will still work! */
	];
	$adminer = new AdminerPlugin($plugins);
	// you can add plugins after instantiating the adminer object 
	// in this way, they can use the referenced object!
	$adminer->addPlugin(new AdminerFormWrapper($adminer));
	return $adminer;
}
// Include original Adminer or Adminer Editor
include './adminer-4.6.3.php';
```

the propagation setting prevents recursively calling plugins.

### DIFF
--- a/plugins/plugin.php
+++ b/plugins/plugin.php
@@ -7,6 +7,13 @@
 * @license https://www.gnu.org/licenses/gpl-2.0.html GNU General Public License, version 2 (one or other)
 */
 class AdminerPlugin extends Adminer {
+	/**
+	 * propagation allows plugins to run and call parent methods from a reference to the adminer instance
+	 * this is useful for doing things in a plugin before AND/OR after calling adminer function
+	 *
+	 * @access protected
+	 */
+	var $propagation=true;
 	/** @access protected */
 	var $plugins;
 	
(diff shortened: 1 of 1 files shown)


Artistan:
this is not quite right yet.

Artistan:
propagation setting was a bit off, works now.

Artistan:
In that case, there are multiple plugins that currently are not functional due to them not being able to

# Example 

https://github.com/natanfelles/adminer-bootstrap-like

## ISSUE - cannot call adminer from plugin.

https://github.com/natanfelles/adminer-bootstrap-like/pull/5

PR #306, opened 2018-10-20
PostgreSQL: Support for identity columns

sartor:
Support for GENERATED BY DEFAULT AS IDENTITY (since 10.0 PostgreSQL version)
See: https://blog.2ndquadrant.com/postgresql-10-identity-columns/


vrana:
Merged.

PR #307, opened 2018-10-25
Horizontal scrollable table.

kPherox:
When I displayed a wide-width table on a small display, I was dissatisfied with the fact that the footer and SQL form were scrolled horizontally and hidden outside the screen. This is a change to resolve this dissatisfaction.

### DIFF
--- a/adminer/create.inc.php
+++ b/adminer/create.inc.php
@@ -170,6 +170,7 @@ foreach ($engines as $engine) {
 <?php } ?>
 
 <?php if (support("columns")) { ?>
+<div style="overflow-x: auto;">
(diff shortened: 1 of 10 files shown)


vrana:
This is pretty nice, I like it. The issue is that the horizontal scrollbar is not visible until you scroll down to the bottom of the table. Thus it may not be obvious that the table is scrollable and scrolling is harder if the mouse doesn't support horizontal scrolling (some mouses don't).

Also please use a class name instead of inline style in widespread changes like this.

I like the feature but could it be solved without the issues? E.g. couldn't `position: sticky` on some elements help?

PR #308, opened 2018-11-17
Horizontal ellipsis

peterpp:
Change three dots '...' to horizontal ellipsis character, which is more suitable for shortened texts.


peterpp:
Done.

PR #309, opened 2018-11-23
<body> class for project-specific styles

lucas-sandery:
Enable editor- or adminer-specific styles by adding an extra class to the body tag.

### DIFF
--- a/adminer/include/design.inc.php
+++ b/adminer/include/design.inc.php
@@ -32,7 +32,7 @@ function page_header($title, $error = "", $breadcrumb = array(), $title2 = "") {
 <?php } ?>
 <?php } ?>
 
-<body class="<?php echo lang('ltr'); ?> nojs">
+<body class="<?php echo lang('ltr'); ?> nojs <?php echo $GLOBALS['project']; ?>">
 <?php
 	$filename = get_temp_dir() . "/adminer.version";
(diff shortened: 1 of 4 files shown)


vrana:
There's nothing special about Adminer or Adminer Editor, plugins or customizations can alter the UI too.

lucas-sandery:
Closing this without any chance for me to reply was a bit of a knee-jerk reaction.

The markup is reduced in Editor and it'd be far easier, and more readable to others to write CSS selectors that begin with `.editor` or `.adminer` than to write a rule that targets one and "undo"/override it for the other. I have never bothered to make sure my theme works well with Editor simply because you don't provide an easy way to differentiate it.

lucas-sandery:
I've updated the branch. Please consider re-opening this PR.

vrana:
Adminer never used `$GLOBALS` and it doesn't use `global` anymore either.

lucas-sandery:
[The updated branch](https://github.com/vrana/adminer/compare/master...lucas-sandery:adminer:project-class) uses neither. It's not showing here, automatically, because the PR is still closed.

vrana:
I've added the classes. I just hope that I wouldn't regret it in the future because people start depending on a stuff like this in very weird ways.

lucas-sandery:
Cheers.

PR #311, opened 2018-11-23
dump_headers incorrectly works with empty $output

petrknap:
Fixed case when the `$output` variable is empty. Original version adds dot at the end of file name, f.e.: `file.ext.`


vrana:
I don't see a case where this could be empty.

petrknap:
We are using Adminer in Symfony 3 application where Symfony does it's "routing + security magic" and then requires Adminer. In our scenario the `$_POST["output"]` is `null` every time and generated files have wrong names.

We can hard code `$_POST["output"] = 'file';` above the require line, but this little change can fix it for everyone and doesn't affect other cases.

PR #312, opened 2018-11-25
Allow responsive styles on larger devices

lucas-sandery:
Currently, resizing a window on a desktop device squashes and truncates content, making the interface difficult to use. E.g. snapping Adminer to a half screen width in Windows to view database content beside something related is more difficult than it needs to be.

There's no advantage to targetting device widths unless dealing with issues in specific browsers.


vrana:
I'm not interested in this change, the user might want to display desktop version in a small window.

lucas-sandery:
🤦‍♂️ I've tried to use the desktop version in a small window; it's awful. One has to scroll about the page *much more* than you would a mobile version. At least Home, End, Page Up, and Page Down make short work of navigating a mobile version on desktop. I don't think I've ever seen a Page Right key.

peterpp:
You can try this responsive theme: https://github.com/pematon/adminer-theme
Maybe it will help.

lucas-sandery:
Thanks, @peterpp. I author [this design](https://github.com/vrana/adminer/blob/master/designs/lucas-sandery/adminer.css) but am frustrated by the use of `{min|max}-device-width` in media queries, rather than `{min|max}-width`, when the latter is generally regarded as best practice.

peterpp:
Oh, I see. And I agree. However, if @vrana is not interested in changing it, you can just ignore it. There is no need to use `max-device-width` in your theme. Just use `max-width` and if you need, you can also add this part from [default.css](https://github.com/vrana/adminer/blob/master/adminer/static/default.css):

```css
@media all and (max-width: 880px) {
	.pages { left: auto; }
	#menu { position: static; width: auto; }
	#content { margin-left: 10px; }
	#lang { position: static; border-top: 1px solid #999; }
	#breadcrumb { left: auto; }
	.rtl .pages { right: auto; }
	.rtl #content { margin-right: 10px; }
	.rtl #breadcrumb { right: auto; }
}
```

Roy-Orbison:
@lucas-sandery Hi, I'd like to submit this PR over at the active fork of this project: [AdminerEvo](https://github.com/adminerevo/adminerevo), but it doesn't seem to exist in either tree. Do you still have it? Or would you be interested in doing so?

lucas-sandery:
@Roy-Orbison

[Pushed.](https://github.com/lucas-sandery/adminer/tree/responsive)

You can append `.patch` to pull URLs to get the changeset: https://github.com/vrana/adminer/pull/312.patch

Roy-Orbison:
Thanks. I did try the patching from the commit URL (adding same suffix) but `git am` could not apply it cleanly. I don't think it liked the mixed line endings.

PR #313, opened 2018-11-26
php <5.4 compat: array syntax

helix84:


vrana:
The static files are compressed and they should be decompressed by `adminer/file.inc.php`.

PR #314, opened 2018-11-29
Use <ul> list for logins

peterpp:
Encapsulate list of logins in the same way as list of tables.


designs/brade/adminer.css
```diff
@@ -56,7 +56,7 @@ outline:0;background:url(//www.bradezone.com/random/adminer_logo.gif) no-repeat;
 * html #lang,* html #menu,* html #breadcrumb{position:absolute}
 * html #lang{padding-top:10px;height:30px}
 * html form#form{height:100%}
-#logins a,#tables a{background: none}
+#logins a,#logins a,#tables a{background: none}
```
vrana:
```suggestion
#logins a,#tables a{background: none}
```

PR #317, opened 2018-12-11
Fix broken html

morpheus-87:
This PR fixes the broken html, that the php script produces.


vrana:
The HTML is not broken, all these tags are optional.

morpheus-87:
But what is the problem to add the closing tags?

vrana:
There's no point in doing so. It only increases code size with no benefit. Make your content-adding script smarter by adding the content at the end of the page if there's no `</body>`.

morpheus-87:
It makes the code much more human-readable and the increase of code size is nearly nothing.

PR #322, opened 2019-01-12
Fix mongo authentication

mirmousaviii:
When I want to sign in with correct username and password, I get this message: Database does not support password.
I think this condition is incorrect.
Also, Adminer handles it when the password is empty and return this message: `Adminer does not support accessing a database without a password, more information.`


vrana:
This is by purpose. Read https://www.adminer.org/en/password/.

mirmousaviii:
Thanks, but it's not worked for MongoDB!

xavdep:
Same problem, don't work auth with password for MongoDB,

Thanks

PR #324, opened 2019-01-20
Adminer Authentification problem mongodb WITH password

xavdep:
Hello,

When I'm trying to connect with adminer to a mongoDB database WITH password, i got this error "Database does not support password."

Authentification works when i try to connect a database WITHOUT password but not work with a database with password

My databases are correctly configured because I can connect on it with others softwares.
Thanks to all

PS: Problem can maybe a link with this thread too https://github.com/vrana/adminer/pull/322

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,6 +1,6 @@
 [submodule "jush"]
 	path = externals/jush
-	url = git://jush.git.sourceforge.net/gitroot/jush/jush
+	url = git://git.code.sf.net/p/jush/git
 [submodule "tinymce"]
 	path = externals/tinymce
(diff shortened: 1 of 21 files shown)


vrana:
First, don't abuse pull requests for support requests.

Adminer doesn't support password-less access to databases by default. To support it, implement the `login` [method](https://www.adminer.org/en/extension/) to verify your Adminer password and the `credentials` method to return the empty password.

djmattyg007:
Is there a reason you don't have issues enabled on the github repo? It's incredibly counterintuitive to have to go to sourceforge to report an issue.

PR #325, opened 2019-01-22
Virtual columns

vrana:


vrana:
Thanks for the extra work. This still has some issues:
- Extra newline after Auto Increment columns.
- Not using `html_select()`.
- Not using `lang()`.
- Using `$jush == 'sql'` instead of `support()`.
- Breaks support for MySQL 4.1 (probably not a big deal but we should strip other MySQL < 5 code elsewhere in a separate commit).
- Useless `GENERATED ALWAYS `.
- Some missing semicolons before `?>` (breaks compilation).
- Always trimming `'` from the default value (must be trimmed only [since](https://mariadb.com/kb/en/library/information-schema-columns-table/) MariaDB 10.2.7).
- Missing entry in `changes.txt`.
- Probably the most important issue is that MariaDB reports the generated columns only [since](https://mariadb.com/kb/en/library/information-schema-columns-table/) 10.2.5 which removes the virtual value whenever the column is edited which makes this hardly usable in lower versions.
- Virtual columns are supported by other databases (MS SQL, Oracle, maybe others). Can the support be added for them too?

Has this been tested with routines? The functionality in `editing.inc.php` is used there too and the virtual columns are not supported there.

PR #331, opened 2019-03-09
Mancave themes (refactor) for floating footer

monolithforge:
Fix mancave themes to use new floating footers by moving designs `mancave` into `mancave-v2` and introducing `mancave-v1`... my favorite daily theme!  (Thanks vrana for the `designs.php` plugin... you rock!)


vrana:
This doesn't merge anymore, please create a new pull request. Also keep `mancave` in the same directory and use perhaps `mancave-hever` for the one with icons. `v1` and `v2` is very confusing. Why would anyone want to use `v1` if there's `v2`?

panreach:
Duly noted!!!

I've been using your adminer.php script for over ten years and absolutely love it.  I easily spend 8 hours per day with an open adminer tab while crunching data on various databases. ;-)

Jeez... I'm a fan... are there `adminer` coffee cups available, because I would totally buy one!

So, because I've been using adminer this long, the mancave-v1 css file is what I've always used... this is why it's so important to me ;-)  I can get stuff done quickly with it... ie. scrolling fast and such.  V2 which was another stolen css theme just isn't as quick when I need to work with data from a UI perspective.

I'm playing with a fresh master and will be work on refactoring the AdminerDesigns Plugin while trying to keep up with your coding style.  I need a darker theme as this seems to be the latest fad. 
 I'd really like to start playing with a new fresh theme called adminer-toor (for the heck of it) while combining both mancave styles while pushing the `Designs` narrative at the same time.

I believe having the `Designs` more prominent  would generate more usage for users in order to compete with all of the other db/view options out there.  Different people like different perspective/views.

Honestly, `adminer` should create a prod/build website that compiles on-the-fly selected options for download.  This would be similar to datatables.net, which is another company that I respect highly.

Does Adminer have a CDN for `Designs`?  The reason I ask is because Adminer is built from the core to be small.  Building the `AdminerDesigns` plugin is lost if there isn't a CDN (unless you would want to compile themes into the production script for developers.  The average user doesn't do this.)

Anyways, Cheers!

I'll try to get a pull/merge request in the next weeks.

Thanks again,
panreach@gmail.com

panreach:
CDN possibility? https://gomakethings.com/how-to-turn-any-github-repo-into-a-cdn/

PR #332, opened 2019-03-15
Keep font size in all tables.

ib4:
Prevent CSS rule "table { font-size: 90% }" from making nested tables
unreadable.


vrana:
Merged, thanks.

PR #333, opened 2019-03-23
Elastic driver - number of records should be count on provided rows

kambo-1st:
Number of the lines in elastic driver should be count on provided lines, not on empty object rows. 
This is solving following warning when you are using elastic driver in PHP 7.3:

_Warning: count(): Parameter must be an array or an object that implements Countable in /var/www/html/adminer.php on line 1228_


vrana:
Merged, thanks.

PR #334, opened 2019-03-23
Elastic driver - show executed query as a json

kambo-1st:
Elastic search query which has been executed should be display in JSON, as this a true format in which is send  to the Elastic search.

Example:

Instead of:
__doc/_search:_ Array ( [size] => 50 )_

There should be this: 
__doc/_search: {"size":50}_


vrana:
Merged, thanks.

PR #336, opened 2019-04-01
Always pass current geometry property with GeomFromText property

dstollie:
# What's this
In our database we have store the geometry properties with an SRID (4326) property. In mysql it is possible to pass this property as a second argument with the `GeomFromText` function. 

Adminer does not pass this second argument and thus resets the SRID property when updating. This PR fixes this by respecting the set SRID property on the column updating.

I tested this code on the insert item feature of Adminer and it in the cases I tested this change it doesn't break anything.


vrana:
Merged.

PR #337, opened 2019-04-01
Cache busting for adminer css

jonathonsim:
When you develop a customised adminer.css to skin adminer, one problem you may face is that changes don't apply to your users until they force-refresh or clear their cache, since the adminer.css file may cache out in either the user's local cache or any intermediate caches or CDN's

The standard way round this is to add a hash to the url that changes when the file changes (adminer does this itself with it's own CSS, using the version number).  

This PR computes a simple (fast) CRC32 hash of the file and adds it to the url.  This means the file will cache and you will experience all the performance benefits of that, but when changed the new version will also be downloaded

### DIFF
--- a/adminer/include/adminer.inc.php
+++ b/adminer/include/adminer.inc.php
@@ -108,7 +108,7 @@ class Adminer {
 		$return = array();
 		$filename = "adminer.css";
 		if (file_exists($filename)) {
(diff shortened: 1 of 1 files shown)


vrana:
Merged.

vlakoff:
Alternatively, you could use [`filemtime()`](https://www.php.net/manual/en/function.filemtime.php).
Better performance, and seems to be the commonly used method.

PR #339, opened 2019-04-04
Provide default value for the parameter in the __construct function to prevent fatal PHP 7.2 error

KZeni:
I'm using PHP 7.2, and I got a fatal error when trying to use this plugin due to `public function __construct($adminer)` not being provided `$adminer` at all times. As such, giving it a default/fallback value resolved the issue while retaining the existing functionality.


vrana:
Just provide it, it's later used by the plugin.

PR #340, opened 2019-04-05
Check if PDO SSL Attributes are set within config

oksiquatzel:
Check if PDO SSL Attributes are set within config, and only set non empty in PDO Options. otherwise Mysql PDO throws errors if attribute is set and empty.

The Usage is, if you only want to set a ca file, so the client can verify server certs, you don't set the key/cert config


vrana:
Merged. Please use whitespace formatting used in the nearby code in your next pull requests (to any project, not just Adminer).

PR #343, opened 2019-05-16
add AdminerAutoUpdater plugin

sad270:
Adminer auto update plugin who check every day by default (can be changed) if adminer is up to date, and download last version automatically if adminer is outdated

### DIFF
--- a/plugins/auto-updater.php
+++ b/plugins/auto-updater.php
@@ -1,0 +1,84 @@
+<?php
+
+/**Adminer Auto Updater
+ *
+ * You should have write permission on adminer.php
+ *
+ * @link https://www.adminer.org/plugins/#use
+ * @author Sadetdin EYILI, https://github.com/sad270/
+ * @license https://opensource.org/licenses/mit-license.php MIT
+*/
+class AdminerAutoUpdater {
+    /** @var string */
+    const COOKIE_NAME = 'AdminerAutoUpdaterLastCheck';
+
+    /** @var string */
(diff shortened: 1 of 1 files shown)


djmattyg007:
This would require giving the webserver write permissions to the code file, which is a horrible idea.

sad270:
> This would require giving the webserver write permissions to the code file, which is a horrible idea.

Yes, i'm using it only on my local webserver. if you have better idea to auto-update without write permission, i'm interrested.

I can change my comment `You should have write permission on adminer.php` by warning like :
`/!\ WARNING : This plugin require giving write permission to the code file adminer.php, which can be dangerous`

Ciantic:
I like this.

I'm pretty sure most of the PHP installations allow writing to files (e.g. WordPress auto-updater kind of depends on it) even though it is a horrible idea.

If we start to nitpick, I think this should check that you are logged in to some database before trying to CURL updates. Otherwise it can be used to DDOS a server by spamming to check the updates, since only check is cookies.

Not sure if any of these 'concerns' are valid though, since this is just a convienient plugin for developers. None is requiring people to use it right?

vrana:
Cookie is not the right mechanism for caching this. The file is stored at server thus it should be cached on server, not on the client. Consider using `filemtime` or similar.

Also, I would prefer if this is published outside Adminer repo. Let me know once it's done and I'll link it from https://www.adminer.org/en/plugins/#user.

PR #344, opened 2019-05-23
pgsql: fix column comment when creating table

helix84:
This fixes the pgsql driver. When creating a new table, it wasn't possible to add column comments at the same time, because it failed with:
```
CREATE TABLE "table1" (
  "col1" integer NOT NULL
);
COMMENT ON COLUMN ""."col1" IS 'col1 comment';

-- ERROR: zero-length delimited identifier at or near """"
-- LINE 1: COMMENT ON COLUMN ""."col1" IS 'col1 comment'
```

### DIFF
--- a/adminer/drivers/pgsql.inc.php
+++ b/adminer/drivers/pgsql.inc.php
@@ -507,7 +507,7 @@ WHERE table_schema = current_schema() AND table_name = " . q($name))));
 					}
 				}
 				if ($field[0] != "" || $val5 != "") {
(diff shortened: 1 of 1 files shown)


vrana:
Thanks, I've fixed it in a slightly different way.

philippebfr:
Hello,

How Can i unsubscribe to this mailing list ?

I have tried via github without success...

Help !

Philippe Bfr
________________________________

PR #345, opened 2019-06-11
Support for viewing and editing of column comments in mssql driver

emsy:


vrana:
Thanks. I've simplified this a little bit. Please make sure it still works.

PR #346, opened 2019-06-18
Fix the SQL command button after updating a record

simon-thorpe:
After updating a record and clicking "Save and continue edit" the green success message displays "Item has been updated. 08:25:00 SQL command".
The button here though doesn't trigger.

This patch fixes this button so it will show the SQL command used to execute the update as expected.


helix84:
I can confirm the problem and the fix on Chrome 74, adminer master.

Another problem on the same screen I noticed is that the "Save and continue edit" button doesn't work, either (edit - "Save and continue edit" updates the value correctly, then the second "Save and continue edit" doesn't update the value).

vrana:
Thanks. Calling it in `ajax()` is too aggressive, I've moved it to the callback in `ajaxForm()` and limited it to `#ajaxstatus`.

PR #347, opened 2019-06-21
PostgreSQL: remove WITH OIDS support

helix84:


vrana:
https://www.postgresql.org/message-id/E1gPG3j-0003cy-03%40gemulon.postgresql.org says that `WITH (oids = false)` is still supported. What issue is this solving?

PR #348, opened 2019-07-05
fix bug #675 mysql composite foreign key constraint

peterdd:
- This replaces the bogus pattern matching with 'show create table' output with just asking information_schema for the required infos.

Please review, I only tested it with 10.3.15-MariaDB-1 Debian 10 yet.

- It also has 2 places where sql queries und their times of the current page request are collected in an array. Hope it does not hurt as it help seeing the bottlenecks of adminer.

### DIFF
--- a/adminer/drivers/mysql.inc.php
+++ b/adminer/drivers/mysql.inc.php
@@ -44,11 +44,14 @@ if (!defined("DRIVER")) {
 			}
 
 			function result($query, $field = 0) {
+				$start=microtime(true);
 				$result = $this->query($query);
 				if (!$result) {
 					return false;
 				}
 				$row = $result->fetch_array();
+				$end=microtime(true);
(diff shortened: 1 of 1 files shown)


vrana:
I didn't notice this before 36ade4e1. Can you please review it?

Next time, avoid adding unrelated features in the same commit.

vrana:
Also delete rather than commenting out old code.

peterdd:
> Also delete rather than commenting out old code.
I haven't evaluated since and if information_schema is available in all mysql versions so wanted keep as alternative.

> Next time, avoid adding unrelated features in the same commit.
Yea, sorry, I know it sucks for reviewers.
Wonder how I will ever get my current schema rewrite accepted when finished. Probably double work to shift everything at the right place. But I'm at least doing good progress with the experiment.

PR #349, opened 2019-07-18
close ul-tag in navigation

peterdd:
always close the ul tag to avoid nesting errors, also added \n for readability of resulting html

### DIFF
--- a/adminer/include/adminer.inc.php
+++ b/adminer/include/adminer.inc.php
@@ -927,25 +927,23 @@ class Adminer {
 </h1>
 <?php
 		if ($missing == "auth") {
-			$first = true;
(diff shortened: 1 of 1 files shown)


peterdd:
changed as requested. please review.

vrana:
There's no need to close `</li>`, it's optional in HTML. Also there's a wrong space in your code. Fixed by d71cb127.

peterdd:
I prefer proper closing, even if HTML5 allows such laziness. It helps debugging for me keeping it xhtml style.

vrana:
One of Adminer's priorities is code size but out of curiosity, how does it help with debugging? What's the difference between `<li>abc</li><li>` and `<li>abc<li>`?

PR #351, opened 2019-07-19
pgsql export constraints

helix84:
Adds features to export in the pgsql driver.

Partially fixes [bug#587](https://sourceforge.net/p/adminer/bugs-and-features/587/).

- export constraints (except TYPEs, to be added)
- first create all tables, only then create all FK constraints (avoids reordering CREATE TABLE statements on order of their FK dependency)

### DIFF
--- a/adminer/drivers/pgsql.inc.php
+++ b/adminer/drivers/pgsql.inc.php
@@ -436,6 +436,23 @@ ORDER BY conkey, conname") as $row) {
 		return $return;
 	}
 
+	function constraints($table) {
+		global $on_actions;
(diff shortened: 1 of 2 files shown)


JanTvrdik:
This would be incredible useful for me as this is by far the biggest issue I encounter when working with Adminer on daily basis.

helix84:
IMO this can be merged. Even if I didn't have time to add support for types, this is an improvement that works as is.

vrana:
Good work. Merged, thanks.

JanTvrdik:
❤️

PR #352, opened 2019-08-01
Enhance AdminerDesigns Plugin to accept limited arguments by theme or no arguments for all themes.

panreach:


panreach:
Gonna try this a different way...

PR #353, opened 2019-08-02
Fix/Enhance AdminerDesigns Plugin to work intuitively.

panreach:


peterdd:
Yeah, and Adminer should have something like a base.css CSS, where the basic css layout is defined.

And get rid of that 90% font size definition.

Themes can then just contain the information of their own custom colors in its simplest case.
More complex themes still can overrule every rule defined in a base.css if they wish.

This avoids redundancy within the themes and also easier to maintain a theme.

This also makes a cache busting recalculation for adminer.css  ( https://github.com/vrana/adminer/commit/7f32e2675941bd16f6d2f3c83df1a8c4ed970834 ) a seldom required case if it is for example just /themename/style.css

plugins/designs.php
```diff
@@ -27,15 +35,18 @@ function headers() {
...
+			$this->designs = array();
+		}
+		if (in_array($_SESSION["design"], $this->designs)) {
+			$return[] = "../designs/".$_SESSION["design"]."/adminer.css";
 		}
 		return $return;
 	}
 	
 	function navigation($missing) {
 		echo "<form action='' method='post' style='position: fixed; bottom: .5em; right: .5em;'>";
```
peterdd: The hard coded style=.. should be removed too.

```suggestion
		echo "<form action='' method='post'>";
```

and go to a base.css / adminer.css

panreach: I would actually agree with this (maybe?), but would you have a preferred location/placement?  If you look at the full stack renditions, the language dropdown is defaulted to the top/left... while logout is defaulted to the top/right.  51% of me loves that @vrana placed it down on the bottom/right... because it actually works really well with most designs.  Sure one could default it to being left of the logout button, but it's current position is growing on me 👍.  It's almost as if it's getting a designated spot in master (which would be awesome if we could figure out how to do compiled theme builds that @vrana would approve of).  With that said, I think it's important to look at all of the production themes as a whole while incorporating this feature.  A lot of those old designs are something I look forward to testing with when any new design change occurs.  For instance, the new floating footer... this should be looked at collectively to make sure all of the production designs work well... well, just because it's fun to have options while maintaining adminer.org as a whole!  Don't get me wrong, if one wants to customize Adminer, it's pretty simple to do, but to make adminer truly compete with phpMyAdmin, I think designs/themes would really help.

I think designs get people involved.  I've been using adminer over phpmyadmin for years.  I always install adminer on each server that I work with using my own customizations.  I started working with a couple of new fresh programmers over the past year, and they won't even touch phpmyadmin now... lol

adminer just freaking works.

`rant/praise finished!`

-Troy

peterdd: Just give that form-tag an id like id="langswitch" and put that to the style(s):

`#langswitch { style='position: fixed; bottom: .5em; right: .5em; }`

This makes also themes easy to change the position of the language switcher, like removing the position:fixed or moving the position somewhere else.

peterdd: Not sure if I got your point of 'rant'. :-)
Adminer still has too many hardcoded style='' attributes in their php functions.

peterdd: I meant #themeswitch but its similiar for #langswitch I think.

peterdd: or #themeselect  or #langselect as Adminer uses html selects for that.

panreach: I believe adding styles to php code really boils down to file size.  @vrana loves making things small.  So, if you could suggest smaller file sizes within the base production code, I'm sure he'd be all ears!

panreach: I'll also say this, the `hever` design rocks them all!!!

peterdd: just put the hardcoded style=''  to adminer/static/default.css

Yeah I took a look at themes too and found only 2 or 3 ok for working with.

panreach: Speaking of themes... I just asked @fat to create a bs5 theme... lol (https://twitter.com/panreach/status/1162157727369564162)

Not sure if he would even respond?

-Troy

vrana:
I don't see a point in this change, it is already demoed in `adminer/designs.php`.

vrana:
> Yeah, and Adminer should have something like a base.css CSS, where the basic css layout is defined.

It already has it in `static/default.css` which is always used even if there's `adminer.css`.

peterdd:
> It already has it in static/default.css
Yes, overlooked it.

PR #358, opened 2019-08-15
setup Mongo authSource by ENV

sinacek:
Enable setup Mongo connection option [authSource](https://www.php.net/manual/en/mongoclient.construct.php) by environment variable.

It's required when credentials are stored in separe collection. 

ENV variable is easiest way for use by docker.

### DIFF
--- a/adminer/drivers/mongo.inc.php
+++ b/adminer/drivers/mongo.inc.php
@@ -618,6 +618,9 @@ if (isset($_GET["mongo"])) {
 		if ($db != "") {
 			$options["db"] = $db;
 		}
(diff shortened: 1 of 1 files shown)


vrana:
This is too magic. You can try implementing it as a plugin.

What's the difference between `db` and `authSource`?

sinacek:
`db` is database with data. `authSource` is database which contains users and their credentials and is used for authentization.
We have more `db` on same machine with one `authSource` for all of them. 

I don't know how to send the information from extension to connection, when there are no arguments. https://github.com/vrana/adminer/blob/master/adminer/drivers/mongo.inc.php#L622

vrana:
OK, merged.

PR #359, opened 2019-08-16
Bug Fix: Generated columns and virtual columns in Mysql broke editing and adding rows

malteriesch:


vrana:
Thanks, merged.

adminer/drivers/mysql.inc.php
```diff
@@ -555,6 +555,7 @@ function fields($table) {
 				"privileges" => array_flip(preg_split('~, *~', $row["Privileges"])),
 				"comment" => $row["Comment"],
 				"primary" => ($row["Key"] == "PRI"),
+				"generated" => in_array($row["Extra"], array("STORED GENERATED", "VIRTUAL GENERATED"))
```
vrana: https://dev.mysql.com/doc/refman/8.0/en/show-columns.html says that this could be either `VIRTUAL GENERATED` or `VIRTUAL STORED`. Is the documentation wrong or is your code wrong?

malteriesch: Now, that is interesting! It seems the documentation might be wrong, just tried again both a virtual and stored generated column on a mysql 8.0.16 server and "STORED GENERATED", "VIRTUAL GENERATED" is indeed what I got respectively for "SHOW FULL COLUMNS FROM <table>"

vrana: Please report a doc bug at https://bugs.mysql.com/.

PR #360, opened 2019-08-17
avoid horizontal scrollbar on variables page

peterdd:
some server variables can get very long without any whitespace or hints for automatic wrapping.


vrana:
I'm not interested in this change.

PR #362, opened 2019-09-02
Update design rmSOFT - version 4.7.3.1 - multicolor (light and dark)

mesaros:


vrana:
Please keep the structure `designs/_name_/adminer.css`. Also choose up to two best designs and stick with them.

PR #363, opened 2019-09-11
Update to design

lucas-sandery:


misfit98:
Yeah. Was about to do same. That width 100% is missing hard. :D

PR #366, opened 2019-10-09
Clickhouse driver fixes

morozovsk:
fixed: 
SQL command for insert, create table, etc. 
before:
screen435
screen438

after:
screen436
screen437

new:
create/rename table
add/drop/change type for column

before:
screen434


morozovsk:
Merge it please.

vrana:
Merged, thanks.

PR #368, opened 2019-11-06
Fix missing file when running in dev mode

simon-thorpe:
The file editing.js would show a 404 console error when running in dev mode and you had /adminer in the path. This keeps it consistent with the functions.js file above it.


vrana:
`editing.js` is different for Adminer and Adminer Editor, hence the relative path.

PR #370, opened 2019-12-12
High-DPI layout

finwo:
On a high-DPI machine, using firefox, input and select elements are way smaller than other elements (see attached image). While this is most likely a bug in the browser in combination with invalid scaling settings in gtk, fixing it shouldn't break behavior for other setups.

2019-12-12-110753_472x171_scrot


vrana:
`font-size` is inherited by default. I'm not interested in mitigating browser bugs in Adminer, please report this to Firefox instead.

finwo:
From what I gather searching around the web, Firefox, Chrome and Safari all have the styling of form elements set to the system theme by default, not the page's. These 3 browsers do not have `font-size` set to `inherit` for form entry elements, requiring the page's styling to set it explicitly.

PR #371, opened 2020-01-10
Added unserialize()

icyz:
I had not realized, sorry for no comment.
This will allow you to see a table of serialized content (like now with json)


vrana:
This should be better as a separate plugin. It's already available at https://gist.github.com/donwilson/0bc0ec7c3701fb20747777a1a7b4cab4.

PR #372, opened 2020-01-16
Plugin login-ip wrong string comparison

Holicz:
Function `strncasecmp` returns `0` when two strings are identical. So to go inside the if branch when two string are identical the `strncasecmp` result needs to be negated.


djmattyg007:
It's probably safer to check explicitly for `=== 0`

vrana:
Shame on me. I remember testing this but it was probably with `127.0.0.1` as allowed IP and `::1` as real IP which is why it passed. Thanks for the fix.

PR #376, opened 2020-03-18
MSSQL comments fixes

emsy:
MSSQL: Don't trucate displayed comments to 50 chars (use varchar maximum of 8000 upon casting)
MSSQL: Allow for multiline comments in alter table (use textarea instead
of input type text)


emsy:
Correcting myself, that should better be VARCHAR(MAX) instead of VARCHAR(8000)

vrana:
I've merged the cast fix. I didn't merge the textarea because it has other consequences (unable to send the form with `Enter`, line wrapping in other places).

PR #378, opened 2020-03-27
add auto-login plugin

Apres2707:
There was a need to provide some users with access to several database servers. At the same time, it was not possible to transfer all of them a login / password. It was also necessary to retain the ability to regulate the list of available servers and user rights. Adminer did an excellent job, we just had to hide the implementation in a small file-plugin.

### DIFF
--- a/plugins/auto-login.php
+++ b/plugins/auto-login.php
@@ -1,0 +1,87 @@
+<?php
+
+/** Docker use example:
+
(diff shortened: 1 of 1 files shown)


vrana:
Please put this into your repo (or a Gist) and I'll link it from https://www.adminer.org/en/plugins/#user.

PR #380, opened 2020-04-02
do not change boolean false to `null` when editing field

CatalinH:


vrana:
I'm not sure what are you trying to fix by this. Please provide repro instructions.

vrana (changes requested):
I'm not sure what are you trying to fix by this. Please provide repro instructions.

CatalinH:
@vrana : 
1. On a postgres database create a table with a nullable boolean column.
`-- Adminer 4.7.7 PostgreSQL dump

DROP TABLE IF EXISTS "test";
DROP SEQUENCE IF EXISTS test_id_seq;
CREATE SEQUENCE test_id_seq INCREMENT 1 MINVALUE 1 MAXVALUE 2147483647 START  CACHE 1;

CREATE TABLE "public"."test" (
    "id" integer DEFAULT nextval('test_id_seq') NOT NULL,
    "test_col" boolean,
    "other_col" character varying NOT NULL,
    CONSTRAINT "test_id" PRIMARY KEY ("id")
) WITH (oids = false);

INSERT INTO "test" ("id", "test_col", "other_col") VALUES
(1,	'1',	'a');

-- 2021-02-08 07:19:33.302825+00`
2. Click to edit the row and uncheck the checkbox (basically set it to false by using the interface)
3. Hit save.
4. Click again to edit the same row, but now (as a real test scenario) to edit other columns (other_col). 
5. Note that the boolean column now is set to null and now, by saving the entry, the value in db will also be updated with null for test_col instead of remaining false/0.

vrana:
I couldn't reproduce it. This is also likely not the correct fix as it would disallow using functions for boolean fields.

CatalinH:
@vrana : I can agree with you it might not be the best approach and I agree that might break some functions, but which functions can be applied on boolean fields? See the issue here: https://youtu.be/Y9dIN0Ib9-c

vrana:
This is reproducible only with PDO_PgSQL, I'll fix it somehow but not like this. Thanks for reporting.

2114L3:
i am on adminer:4.8.0-standalone, postgres 13.2-alpine, PDO_PgSQL and still getting 
`ERROR: invalid input syntax for type boolean: ""`
as soon as i hit edit on a row with a boolean and screen full of php warnings cause false is going to an invalid value
hitting save also shows complaint about the boolean being blank in the WHERE clase

vrana:
Fixed, thanks.

PR #382, opened 2020-04-05
BLOB QoL on sqlite

L-P:
1. If you have UUIDs stored as BLOB and used as foreign keys, Adminer would
create a link to download the binary version of the UUID rather than
linking to the FK destination.

2. Implement hex/unhex for sqlite.

### DIFF
--- a/adminer/drivers/sqlite.inc.php
+++ b/adminer/drivers/sqlite.inc.php
@@ -774,10 +774,17 @@ if (isset($_GET["sqlite"]) || isset($_GET["sqlite2"])) {
 	}
 
 	function convert_field($field) {
+		if ($field["type"] === 'blob') {
+			return "HEX(" . idf_escape($field["field"]) . ")";
+		}
 	}
 
 	function unconvert_field($field, $return) {
-		return $return;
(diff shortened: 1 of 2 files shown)


vrana:
1. Downloading the blob seems more important than linking the other table.
2. MySQL uses `HEX` for `binary` (short binary fields), not for `blob`.

L-P:
I think you may have misunderstood what this PR is about @vrana 

1. Unless you're doing something icky, I see no reason to have a FK as blob
unless it's an UUID, thus, linking to the corresponding entry in the other
table seems to be the sensible (and expected, for me) thing to do. What good
does downloading an UUID does ?  
I could ensure the field is 16 bytes wide before linking if you want this to
happen _only_ for this the case of UUIDs, but it could seem a bit arbitrary
(and I don't like special cases).

2. This is specific to SQLite, I don't understand your remark on MySQL. Is it about my changes in `select.inc.php`?

vrana:
2. MySQL has two distinct types - `blob` and `binary`. Adminer uses `HEX` only with `binary`. SQLite doesn't have such distinction and using `HEX` for everything seems too much.

PR #385, opened 2020-04-26
Added Redoracle theme design

redoracle:
my first test


vrana:
The dark gray on the dark blue is hard to read. There are some empty boxes for an unknown reason.
image

vrana (changes requested):
The dark gray on the dark blue is hard to read. There are some empty boxes for an unknown reason.
image

PR #386, opened 2020-04-29
Pgsql generated always fix

sartor:
Fix bug where columns with GENERATED ALWAYS BY IDENTITY columns were without any information in table structure mode in Postgresql 10+


vrana:
This breaks PostgreSQL < 10. Please be more careful next time. Merged with that fixed.

PR #387, opened 2020-05-03
Fixes to lucas-sandery theme

lucas-sandery:
* Work around Edge svg + `mask-image` bug.
* Style file inputs.
* Prevent language & logout buttons disappearing on scroll.
* Correct position of column heading actions
* Long queries overflowed the page
* Some better button focus/hover states


vrana:
Merged, thanks.

PR #388, opened 2020-05-09
use password as $encryption_key to SQLite3::open.

zh4ui:
in order to apply this, a plugin has to customize $adminer->credentials(),
and return "locksafe" as the server name.


vrana (changes requested):
This is pretty cool, I didn't know that SQLite supports encryption. Instead of requiring the users to use a plugin with a magic server name `locksafe`, please detect if [SQLite Encryption Extension](https://www.sqlite.org/see) is available and if it is then just use the password returned by `credentials` to decrypt it.

PR #389, opened 2020-05-14
Just add submodule to link some theme

shiipou:
You have good theme on your website, so why not put it in your design directory ?
I've just add submodule to link some theme I like. In this way, you have a link directly to the github repository of the theme creator.

This allow to use the environment variable ADMINER_DESIGN="hydra" in the docker image.
Very useful.


vrana:
Good idea, thanks. Merged.

PR #390, opened 2020-05-21
Hide sqlite_sequence table in the database schema

FrancoisCapon:
The table `sqlite_sequence` is an "internal table"  (for the AI) of SQLite3.
I think that this table must not figure on the database schema.

Thanks for adminer.


vrana:
I use this table now and then.

PR #392, opened 2020-05-31
PostgreSQL: Fix current schema in view definition

helix84:
Displaying a view definition in Adminer didn't reflect current schema. It listed all oids with the given view name and then displayed the definition of the first such view.


vrana:
Merged, thanks.

PR #393, opened 2020-06-09
show correct collation for PostgresSQL

asmarcz:
Displays correct collation for database in Select database menu.


vrana:
This seems to be supported only since PostgreSQL 8.4. Merged with a version check.

PR #394, opened 2020-06-19
add alternative design dracula theme

douglasdamasio:
Hi,

My name is Douglas, and I would like to submit a new design to the Design Hall that you already have.
Below is an image for illustration.

img

It is called dracula, it has the official color palette of [Dracula Theme](https://draculatheme.com/) and was based on the already existing theme of [Pepa Linha](https://github.com/vrana/adminer/tree/master/designs/pepa-linha).

Hope you like it.

Bye = D


vrana:
Colorful! Added, thanks.

PR #395, opened 2020-07-21
fix wrong hungarian translation

pbence:


adaamz (approved):
according to google translate, there was translation for `Columns` instead of `Rows` :)

vrana:
Merged, thanks.

PR #396, opened 2020-08-11
Postgres - fix missing partitioned tables in navigation

rydercz:
From Postgres version 10 have partitioned tables `relkind = p`.

https://www.postgresql.org/docs/10/catalog-pg-class.html


dark0ni:
Same issue, our price tables are splitted in 20 table entries. It would great if somebody can code review this merge request

salacr:
I think that this MR is incomplete, for example it seams that partitioned tables will be handled as a VIEW. (Well that's actualy not incorect) but when we want to delete paritioned table the syntax is DROP TABLE right ?

vrana:
Merged as is.

PR #397, opened 2020-09-04
Adminer mongodb driver fix bug

dream-mo:
fix bug: log in mongodb error: "Database does not support password", but the user name and password are correct and it is required,and user can not log in successfully. The error message and logic make people feel very strange.

### DIFF
--- a/adminer/drivers/mongo.inc.php
+++ b/adminer/drivers/mongo.inc.php
@@ -624,10 +624,9 @@ if (isset($_GET["mongo"])) {
 		try {
 			$connection->_link = $connection->connect("mongodb://$server", $options);
 			if ($password != "") {
-				$options["password"] = "";
+				$options["password"] = $password;
 				try {
 					$connection->connect("mongodb://$server", $options);
-					return lang('Database does not support password.');
 				} catch (Exception $ex) {
 					// this is what we want
 				}


morozovsk:
>The error message and logic make people feel very strange.

But the logic is ok:
* adminer connects with password
* if adminer also can connect without password then you have error
* if adminer can't connect without password then everything is ok

```php
$connection->_link = $connection->connect("mongodb://$server", $options); // adminer connects with password:
if ($password != "") {
	$options["password"] = "";
	try {
                // if adminer also can connect without password then you have error
		$connection->connect("mongodb://$server", $options);
		return lang('Database does not support password.');
	} catch (Exception $ex) {
		// this is what we want
                // if adminer can't connect without password then everything is ok
	}
}
```

So you need to setup your mongodb properly.

oderwat:
AFAIK there is no way to set up MongoDB in a way that you can not log in without a password.  as there is an option to authenticate after being connected. I wonder how just anybody can setup MongoDB otherwise? I would appreciate getting this information! Besides that, the current Adminer version does not really work with MongoDB (4.4) for me. Even viewing data of a collection throws an error about wrong BSON types.

realkarmakun:
MongoDB has ability to authenticate after connection, so Adminer "thinking" that MongoDB 
> So you need to setup your mongodb properly.

@morozovsk 
That is not true. MongoDB supports authorization after connection, so currently checks that are made in Adminer is not enough to say that MongoDB instance doesn't have authorization enabled. I haven't found a way to configure it diffrently in MongoDB. According to official [documentation](https://docs.mongodb.com/manual/tutorial/enable-authentication/) setting `--auth` flag on `mongod` startup and creating Administrative User more than enough to enable authorization in MongoDB instance. 
To keep Adminer's way to dealing with passwords there can be done additional checks (since unauthorized connection only allow `db.auth()` as far as I know)

morozovsk:
>To keep Adminer's way to dealing with passwords there can be done additional checks (since unauthorized connection only allow db.auth() as far as I know)

But it is not realized in this pull request. This pull request just disable checks.

realkarmakun:
> But it is not realized in this pull request. This pull request just disable checks.
Yeah sorry about that, just wanted to make sure that contributor know about the issue and I couldn't find a good place to contact you. It's just that Adminer doesn't work with MongoDB atm (but php5 version works fine for some reason)

vrana:
#405 is the right fix.

PR #398, opened 2020-09-08
Fix driver mongodb: Database does not support password.

nujuna:
Fix driver mongodb: Database does not support password.


morozovsk:
https://github.com/vrana/adminer/pull/397#issuecomment-716203427

PR #400, opened 2020-10-06
PostgreSQL (editor): Fix for broken search on timestamp/boolean fields (sf: #325)

mevdschee:
Fix for broken search on timestamp/boolean fields in PostgreSQL for Adminer Editor

See: https://sourceforge.net/p/adminer/bugs-and-features/325/


mevdschee:
Reproduction steps: make a table with timestamp and boolean fields. Use adminer editor (!) search (on any field) functionality. You will see error messages (as reported in [sourceforge issue 325](https://sourceforge.net/p/adminer/bugs-and-features/325/)). This (minor) change fixes the error messages in most cases. The problem still exists when you use the ">=" or "<=" operator in combination with the "any field" option. This is an edge case and occurs less often.

vrana:
I've used `$driver->convertSearch` instead which is driver aware.

mevdschee:
Thank you for the great software. I use it on a daily basis with mysql and thank for fixing this (postgresql) issue. Please keep up the good work!

PR #401, opened 2020-10-20
Various PHP 8 fixes

TimWolla:
This PR fixes various new warnings in PHP 8. There are still quite a few remaining, but I'd like to put this up for review already.

After this PR is merged Adminer requires at least PHP 5.6, because of the spread operator in the `query()` method signature.

### DIFF
--- a/adminer/include/auth.inc.php
+++ b/adminer/include/auth.inc.php
@@ -73,7 +73,7 @@ if ($auth) {
 		redirect(auth_url($vendor, $server, $username, $db));
 	}
 	
-} elseif ($_POST["logout"]) {
+} elseif (isset($_POST["logout"])) {
 	if ($has_token && !verify_token()) {
 		page_header(lang('Logout'), lang('Invalid CSRF token. Send the form again.'));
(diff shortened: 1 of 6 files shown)


djmattyg007:
It would be nice to drop support for older versions of PHP at some point. PHP7 is now five years old.

octalmage (approved):
Looks good! lets get this merged. I tested this with PHP 8 and it resolved the fatal errors I saw.

vrana:
The `isset` calls are completely useless. The rest is addressed by #406.

vrana:
Thanks for the list of possible unset array keys, I'll use in other way.

vrana:
Actually, this is vastly incomplete.

TimWolla:
Yes, I am aware as indicated in my initial message:

> There are still quite a few remaining, but I'd like to put this up for review already.

I am not familiar with the code base and I hoped at least doing some of the necessary changes would you help here. Use my commits / pull request as you like. I do not plan an making any further changes to it. Thanks.

PR #402, opened 2020-10-28
Fix displaying type mapping for Elasticsearch >= 6.0

peterpp:
Earlier versions of Elasticsearch (<= 5) supported multiple types per index. That meant that you could have different data mappings for each type. With Elasticsearch 6, this was removed and you can only have single mapping type.


vrana:
Thanks, merged with using `min_version(6)`.

PR #403, opened 2020-10-30
Add optional color for connections

essembeh:
Adminer has become a tool I use very often recently and I have many tabs connected to many databases at the same time.
I wanted to have a way to quickly distinguish databases to avoid manual errors, so I implemented an optional color you can choose at login time (and add to bookmarks `&color=red`), this helps me a lot, maybe this could help others?


vrana:
Try to implement this as a plugin. Avoid doing unrelated changes (such as adding `</option>` which is optional in HTML).

PR #404, opened 2020-11-12
Make the ME constant pluggable

Roy-Orbison:
When creating an integration of Adminer as a third party service, in a similar manner to PHPMyAdmin, query string parameters used for selection/authentication are destroyed. This hinders the ability to integrate Adminer when robust, non-rewritten URLs are used.

For example, I try to make Adminer accessible at `https://example.com/admin-ui/3rd-party.php?3ptask=adminer&3ptoken=abc123` but all generated URLs will have those parameters removed like so `https://example.com/admin-ui/3rd-party.php?file=default.css&version=x.y.z` which would cause failure.

This constant enables preservation of parameters via a simple define:

```php
define('MEQPFX', '3ptask=adminer&3ptoken=' . rawurlencode($_GET['3ptoken']) . '&');
```

### DIFF
--- a/adminer/include/bootstrap.inc.php
+++ b/adminer/include/bootstrap.inc.php
@@ -84,7 +84,11 @@ include "../adminer/drivers/mysql.inc.php"; // must be included as last driver
 
 define("SERVER", $_GET[DRIVER]); // read from pgsql=localhost
 define("DB", $_GET["db"]); // for the sake of speed and size
+if (!defined('MEQPFX')) {
+	define('MEQPFX', '');
+}
 define("ME", str_replace(":", "%3a", preg_replace('~\?.*~', '', relative_uri())) . '?'
+	. MEQPFX
 	. (sid() ? SID . '&' : '')
 	. (SERVER !== null ? DRIVER . "=" . urlencode(SERVER) . '&' : '')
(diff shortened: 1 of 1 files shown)


vrana:
I see your issue but I'm not going to accept enabling this through some magic constant. The easiest solution is probably to use URL rewriting.

Roy-Orbison:
>The easiest solution is probably to use URL rewriting.

It really isn't. URL rewriting happens on the way *in*, this changes URLs on the way *out*. `ME` is used *everywhere* and you've ensured there's no way to control it. Rewriting output is a hack because it could be in another encoding, e.g. JSON.

If you hadn't made `ME` a constant, I wouldn't have suggested this solution. There's no other place to change it properly, so if a "magic" constant isn't good enough, what is?

vrana:
Access your script as `3rd-party-adminer-abc123.php` and rewrite it to what you need. Adminer will still see the original URL.

Roy-Orbison:
As I wrote in my original report, this is about **non-rewritten** URLs. There's a good reason why many admin interfaces deliberately avoid rewriting, instead use a plain file hierarchy and query strings. It eliminates a significant, unnecessary point of failure.

PR #405, opened 2020-11-24
Fix MongoDB connect 'Database does not support password.' tips

imdong:
Add MongoDB Drivers func ping.

According to the official document: [MongoDB\Driver\Manager::__construct](https://www.php.net/manual/en/mongodb-driver-manager.construct.php)

> Note: Per the » Server Discovery and Monitoring Specification, this constructor performs no I/O. Connections will be initialized on demand, when the first operation is executed.

After creating the object, perform any operation to generate I/O connection to verify that the password is correct.

Do not add `ping` validation to the correct password section to reduce changes to unknown code.

> Code reference from [document Example #1](https://www.php.net/manual/en/mongodb-driver-manager.executecommand.php#refsect1-mongodb-driver-manager.executecommand-examples)

### DIFF
--- a/adminer/drivers/mongo.inc.php
+++ b/adminer/drivers/mongo.inc.php
@@ -194,6 +194,10 @@ if (isset($_GET["mongo"])) {
 			return $connection->_db->selectCollection($_GET["select"])->count($where);
 		}
 
+        function ping($link) {
+            return true;
+        }
+
 		$operators = array("=");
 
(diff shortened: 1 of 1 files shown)


vrana:
Thank you so much for tracking this bug down and properly fixing it. Several users reported problems here but it didn't surface with the legacy extension. Merged.

PR #408, opened 2020-12-21
Attempt to workaround for MySQL8 error 'Incorrect DATETIME value' WIP

dg:
Allows to search in tables where is a datetime column.

image

### DIFF
--- a/adminer/drivers/mysql.inc.php
+++ b/adminer/drivers/mysql.inc.php
@@ -1061,6 +1061,12 @@ if (!defined("DRIVER")) {
 	* @return string
 	*/
 	function convert_field($field) {
+		if (min_version(8) && ($field["type"] === "datetime" || $field["type"] === "timestamp")) {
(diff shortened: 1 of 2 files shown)


vrana:
I couldn't reproduce this. Please provide repro steps. Perhaps `sql_mode` affects this? You also might want to try modifying `convertSearch` instead of changing `convert_field` and `WHERE` to `HAVING` (which might have performance consequences.

dg:
MySQL: 8.0.22
sql_mode: STRICT_TRANS_TABLES,NO_ENGINE_SUBSTITUTION

This command generates error **Chyba v dotazu (1525): Incorrect DATETIME value: 'xx'**

```sql
SELECT * 
FROM `table`
WHERE (`col` = 'xx')
LIMIT 50
```

vrana:
Thanks. I've fixed it by skipping date columns for non-date values in search anywhere.

PR #409, opened 2021-01-12
fixes Unsupported operand types when $port is socket

dg:


vrana:
This was changed just recently in 35bfaa75 to fix [bug 765](https://sourceforge.net/p/adminer/bugs-and-features/769/). Consider using blame to see why the code is there. I've fixed it hopefully without introducing this regression.

dg:
Fatal error https://3v4l.org/6E3q8

dg:
Thanks

PR #410, opened 2021-01-26
Title for select data link

tiso:
For better accessibility themes with icons & symmetry - next-by link to structure in menu has title.


vrana:
Merged.

PR #411, opened 2021-01-27
Linkify URLs in SQL results

apeschar:
This PR turns a column value that looks like an URL into a link, just like on the "Select data" page.


vrana:
Merged, thanks.

PR #412, opened 2021-02-08
Bug in query for getting triggers on a table which missed schema condition

salacr:
Fixing query for getting triggers on a table which missed schema condition


vrana:
Merged, thanks.

PR #414, opened 2021-02-16
Delay adminer object init to after constants have been set.

icedream:
The update to 4.8.0 includes 4742bde873598e3017c566777b37e281996514b5, which causes `adminer_object` to be called before the MySQL driver is initialized and before any of the Adminer constants have been initialized. This can cause external plugin code to fail if it relies on these constants in order to set up things before Adminer takes any authentication-related actions (e. g. via a plugin-provided constructor function).

This PR changes the order to be more compatible and allow these plugins to continue to work.

NOTE: I am currently marking this as draft as I still have to verify whether this is enough to fix the issue without introducing negative side effects.


vrana:
I've done this to support drivers in plugins. Show me your external plugin and I'll try to find another solution.

PR #416, opened 2021-02-25
Prevent adminer to consider MariaDB sequences being views

vlada-dudr:
MariaDB 10.3 added support for sequences. This little change prevents Adminer to treat them as views. Sequences are implemented as special tables, so one can work with them little bit (see values, drop them) like that.


vrana:
Merged, thanks.

PR #418, opened 2021-03-12
Use ST_SRID() in mysql driver for MySQL 8

bcremer:
`SRID()` is no longer available in MySQL 8 and has been replaced with `ST_SRID`.

This is a follow-up to: https://github.com/vrana/adminer/commit/327041874ecdb398436adc8a5d32d49c6866dd8c


vrana:
Merged, thanks.

PR #419, opened 2021-03-18
OPT setup

aftechro:
Hello,

can you please provide instructions on how to install and use the OTP?

image

i already downloaded the files but i got stuck on number 3. Create this file

what file should i create? is that the entire code (as per screen shoot)?

Already created test.php place that code, but when i`m accessing in on the browser domain.com/test.php is just an empty blank page

am i missing something?

Thank you!


vrana:
You probably just need to adapt your paths. Don't use pull requests as support forum.

PR #420, opened 2021-03-21
Fix selecting from Elasticsearch with conditions

peterpp:
Query type "filtered" has been depraceted and is unsupported in current Elastic version. "bool" query has to be used instead. This pull request:

- fixes searching
- adds "must", "should", "must_not" conditions
- fixes searching in system "_id" field
- displays real error retuned by Elastic instead of general "Invalid credentials."
- remove empty body from requests that don't support them (7.14 compatibility)


peterpp:
I rebased this branch onto `field-priviledges` (https://github.com/vrana/adminer/pull/422)

cweiske:
Displaying the real error is indeed a problem with 4.8.0. I get "invalid credentials" for every error.

PR #421, opened 2021-03-26
Thank you!

helix84:
This is not an issue, just a public thank you to Jakub Vrána (@vrana) and other developers for making, maintaining and improving Adminer. I included some of my thoughts why the project is valuable to me and where I'd like to see it in the future. If Jakub decides to keep this "issue" open for a while, others may chip in with their thanks and thoughts.

I've been using Adminer since at least 2012 (judging by my first patch). Before that I used phpMyAdmin simply because it has been a popular and widely known tool. I might have been shopping for an equivalent tool for PostgreSQL when I came upon Adminer. What appealed to me apart from support of multiple DBMSes was that I could deploy it as a single PHP file (PHP being ubiquitous then and now, and simple to set up).

I has been in my toolbelt ever since as my main way of writing ad-hoc queries and DBMS management. Adminer's support helped me explore other RDBMses (like Oracle) using a familiar interface.

What I appreciate the most about Adminer is
- its ease of deployment
- that Jakub has been continuously maintaining it for a long time with stable project management quality (releases, changelogs, docs, PR merges, responsiveness to issues, own development, general improvements)
- and its support of many backends (and the modularity that enables this potential to extend it).

The general areas where I see potential for improvement is
- its bus factor of 1 (where the obvious problem is that it's hard to find other highly qualified maintainers)
- marketing (improving market share could improve the chances of the former) and
- testing (CI) on multiple backends.

Thank you, Jakub, for making and maintaining Adminer. Thanks to all fellow contributors for improving it! :clap:

While I haven't donated financially, if you appreciate the project, it might be the easiest way for you to support it (or get your employer to support it):
[Patreon](https://patreon.com/jakubvrana)
[PayPal](https://sourceforge.net/p/adminer/donate/)

### DIFF
--- a/readme.txt
+++ b/readme.txt
@@ -19,3 +19,5 @@ lang.php - Update translations
 tests/katalon.html - Katalon Automation Recorder test suite
 
 If downloaded from Git then run: git submodule update --init
+
+Thank you!


unix4you2:
Thanks from here too!!!

vrana:
Thanks for the kind words. I'm actually not a very good maintainer, I sometimes leave action items for months without a response…

PR #423, opened 2021-04-16
Create LICENSE

gauravojha:
readme.txt mentions `Apache License 2.0 or GPL 2` as license. Creating PR to create a LICENSE file using Apache 2.0


unix4you2:
Hi there.
Should we put a LICENSE file only for Apache?
As I understand Adminer is a dual license software so if we put a file only for Apache what about people that want to use it under GPL 2 terms?
Best regards.

gauravojha:
@unix4you2  thanks for the review. 

followed the conventions suggested by SPDX. 

Links [here](https://events.static.linuxfound.org/sites/events/files/Introduction%20to%20SPDX-without%20graphics.pdf) and [here](https://spdx.github.io/spdx-spec/1-rationale/) 

I can remove individual LICENSE.* files if deemed unnecessary, but if we want to keep it, would need information from @vrana  on these fields  - year and name of author

```
Copyright (C) <year>  <name of author>
```

vrana:
What is this good for? The licenses are publicly available, why should they be included also here?

djmattyg007:
Because the general wisdom is that a software's license should always be included with all copies of the software itself, regardless of what form the software is being transmitted in.

gauravojha:
@vrana  adding to what @djmattyg007  said, adding a License at the top level of your repo also makes it readable by parsers and available as an info on github. Of course it is not absolutely required, but it is generally a good idea for popular projects to expose this information.

At the end of the day, it is your call, but I would really support the idea of including this in your repo. 

Some links:
https://docs.github.com/en/github/creating-cloning-and-archiving-repositories/creating-a-repository-on-github/licensing-a-repository
https://docs.github.com/en/communities/setting-up-your-project-for-healthy-contributions/adding-a-license-to-a-repository
https://swcarpentry.github.io/git-novice/11-licensing/index.html
https://medium.com/glitch/why-add-a-license-to-your-open-source-project-6f07941f9288

vrana:
Added.

PR #426, opened 2021-05-09
multiple inserts() when create new record

sadn3r:
fixed by sending form ajax & disable form.
easy to repeat by simulating a slow connection in dev tools(multiple press of the save button)


djmattyg007:
Disabling a form when the user submits it is a good usability improvement 👍

vrana:
I'm not interested in this change.

PR #427, opened 2021-05-17
add ability to pass sslmode for postgres

wodka:
add ability to force the ssl mode by adding the config for the login-ssl plugin

```php
<?php
require_once('plugins/login-ssl.php');

return new AdminerLoginSsl([
  "sslmode" => "require",
]);
```


kepi:
Is there any possibilty to merge this? We are using this for connecting to external postgre servers, where there is fortunately no possibility to connect without SSL.

toabi:
Oh… jeah… that would be really useful :)

peterpp:
Properly implementted here: https://github.com/pematon/adminer/commit/ef793eddf6a88e1b9c2b277563bbe1ac42bf5d64

(compatibility with existing MySQL options, support for both `pgsql` and `pdo_pgsql` extensions.

vrana:
Merged, thanks.

PR #433, opened 2021-07-28
Customizable import filename

emr-dev:
Hello. I decided to add the ability to select a file when importing data, it would seem a very simple thing, but it saves a lot of time and makes this functionality very useful for everyday work. Thank you!

### DIFF
--- a/adminer/include/adminer.inc.php
+++ b/adminer/include/adminer.inc.php
@@ -908,12 +908,15 @@ class Adminer {
 		return $ext;
 	}
 
-	/** Set the path of the file for webserver load
-	* @return string path of the sql dump file
-	*/
-	function importServerPath() {
-		return "adminer.sql";
-	}
+
+    /**
+     * List of files to import
+     * @return array
+     */
+    function importFiles() {
+        return glob(__DIR__ .'/*.{sql}', GLOB_BRACE);
+    }
+
 
 	/** Print homepage
 	* @return bool whether to print default homepage
--- a/adminer/include/adminer.inc.php
+++ b/adminer/include/adminer.inc.php
@@ -914,7 +914,11 @@ class Adminer {
      * @return array
      */
     function importFiles() {
-        return glob(__DIR__ .'/*.{sql}', GLOB_BRACE);
+        $files = glob(dirname(__FILE__) . '/*.[Ss][Qq][Ll]');
+        if (!is_array($files)) {
+            $files = array();
+        }
+        return $files;
     }
 
 
--- a/adminer/sql.inc.php
+++ b/adminer/sql.inc.php
@@ -237,16 +237,9 @@ if (!isset($_GET["import"])) {
 	$importFiles = $adminer->importFiles();
 	if ($importFiles) {
 		echo "<fieldset><legend>" . lang('From server') . "</legend><div>";
-		$options = null;
-		foreach ($importFiles as $filename) {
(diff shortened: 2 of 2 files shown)


adminer/include/adminer.inc.php
```diff
@@ -908,12 +908,15 @@ function dumpHeaders($identifier, $multi_table = false) {
...
-	function importServerPath() {
-		return "adminer.sql";
-	}
+
+    /**
+     * List of files to import
+     * @return array
+     */
+    function importFiles() {
+        return glob(__DIR__ .'/*.{sql}', GLOB_BRACE);
```
djmattyg007, line 917: It looks like you don't actually need the braces here.

Also, last I checked, musl libc (used by Alpine, which is popular in the docker world) doesn't support `GLOB_BRACE`, so it would definitely be worth avoiding.

emr-dev: Thanks, i changed the function

emr-dev (commented):
edit importFiles function

emr-dev:
> edit importFiles function

Roy-Orbison:
This doesn't validate the POSTed path against the file list. Probably not a big security issue if you can already upload SQL files, but makes sense to check its existence.

Also `glob` has been supported since PHP v4.3, and those braces were a no-op, so you could change it to this to simplify and error-proof it:
```php
$files = glob(__DIR__ . '/*.[Ss][Qq][Ll]');
if (!is_array($files)) { # in case boolean false on error
	$files = array();
}
return $files;
```

adminer/sql.inc.php
```diff
@@ -234,10 +234,19 @@
...
 		echo "<fieldset><legend>" . lang('From server') . "</legend><div>";
-		echo lang('Webserver file %s', "<code>" . h($importServerPath) . "$gz</code>");
+		$options = null;
+		foreach ($importFiles as $filename) {
+            $options .='<option value="'.$filename.'">'.$filename.'</option>';
+        }
+
+		echo '
+		<select name="webfilename" />
+		  '.$options.'
```
Roy-Orbison, line 247: Adminer provides the `optionlist()` function for that.

Roy-Orbison:
Not trying to be picky. I think this is a good feature, and am trying to increase the chance it'll be accepted.

Roy-Orbison:
It turns out [this already exists](https://gist.github.com/joshcangit/ad28f82baf1c9c2fab22dd9e8f39f799) as a plugin, though that one is somewhat broken. I wrote a plugin from scratch before looking for any existing one, which was probably a mistake but maybe better since I didn't start with bugs. [I published my version](https://gist.github.com/Roy-Orbison/52f20d5b2eacd9ad3b828ca225202982) as a fork of the original because credit where it's due, but I think mine is much improved.

I doubt this PR will get accepted because the function your patch deletes is already designated as pluggable, it's already plugged, and your desired behaviour isn't quite there. Thanks for putting me on to this, though.

emr-dev:
Cool, but why such a seemingly simple function cannot exist
standard adminer. Many users use the standard version and do not waste time installing different plugins.

Roy-Orbison:
It's more secure that way. It's not ridiculous to imagine a destructive SQL dump in the same directory as Adminer. Limiting to `adminer.sql[.gz]` by default reduces the potential for damage, makes it the file's purpose clear, and likely reduces obsolete files because you'd typically overwrite the previous import.

PR #434, opened 2021-07-30
PostgreSQL: exclude internal schemas from listing

helix84:
pg_temp_% and pg_toast_temp_% are postgres-internal schemas for temporary tables and there is no reason why a normal user would need to access them. This PR declutters the list of schemas by hiding them.

### DIFF
--- a/adminer/drivers/pgsql.inc.php
+++ b/adminer/drivers/pgsql.inc.php
@@ -710,7 +710,7 @@ AND typelem = 0"
 	}
 
 	function schemas() {
-		return get_vals("SELECT nspname FROM pg_namespace ORDER BY nspname");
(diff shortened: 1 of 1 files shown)


laluka:
Better add it as an option, "some" users really want to have the full content of the db without using the CLI, so hiding fields without telling the user (us), is quite annoying.. :/
I'd rather keep it that way, or add an "hide" option for these tables ! :)

peterpp:
Hiding system schemas and tables will be a simple configuration option in upcoming version 5 of [AdminerNeo](https://github.com/adminerneo/adminerneo) fork. It is already implemented, feel free to check it out 😊

vrana:
Database administrators might be curious about them.

PR #435, opened 2021-08-12
Dark theme for Adminer

bilal-yameen:
# Dark theme for Adminer

### Screenshots
enter image description here

enter image description here


peterpp:
This theme is not compatible with plain Adminer. You are using my theme as the base. It requires additional plugin and resources (https://github.com/pematon/adminer-theme).

bilal-yameen:
@peterpp I've taken out the theme from plugin to work for plane Adminer.
Btw thanks for the base theme though 😊...

martin-braun:
@peterpp Would it be possible to update the default theme to have a dark mode included, so it auto-toggles with the system preference?

peterpp:
> @peterpp Would it be possible to update the default theme to have a dark mode included, so it auto-toggles with the system preference?

It would be possible in my original theme. I'll try to find some time.

martin-braun:
@peterpp Awesome, this would be a killer feature for me, will definitely spare some dimes for pleasing my eyes based on my system preference (I use light and dark depending on the time of the day). Long story short: I would be very grateful. :)

peterpp:
@martin-braun If you are still interested, I am working on [AdminerNeo](https://github.com/adminerneo/adminerneo) fork with redesigned default theme in upcoming version 5. It will support dark mode and configurable color variants.

vrana:
This is what I see.
image

PR #436, opened 2021-08-19
Check unlimited memory limit

oksiquatzel:
usefull if adminer runs in containerd Environment where resources are controlled by the container engine


peterpp:
Merged with small modification to https://github.com/pematon/adminer, thanks.

vrana:
Thanks.

PR #437, opened 2021-08-22
Fix elasticsearch(>=7.11) does not support body in GET request

tauh33dkhan:
Elasticsearch(>=7.11) does not support the body in GET request and returns an error message when making GET request with body such as the login request, For more details read this [elasticsearch issue](https://github.com/elastic/elasticsearch/issues/69532) and the pull request that introduced this change [66043](https://github.com/elastic/elasticsearch/pull/66043).

To fix this I made changes in the `rootQuery` function of elasticsearch driver to only send the request body and content-type header if the request method is not `GET`.

Fix tested on:
* Elasticsearch Version: 7.14.0
* Elasticsearch Version: 7.7.0

### DIFF
--- a/adminer/drivers/elastic.inc.php
+++ b/adminer/drivers/elastic.inc.php
@@ -16,12 +16,17 @@ if (isset($_GET["elastic"])) {
 			 */
 			function rootQuery($path, $content = array(), $method = 'GET') {
 				@ini_set('track_errors', 1); // @ - may be disabled
-				$file = @file_get_contents("$this->_url/" . ltrim($path, '/'), false, stream_context_create(array('http' => array(
+				$opts = array( 'http'=>array(
 					'method' => $method,
-					'content' => $content === null ? $content : json_encode($content),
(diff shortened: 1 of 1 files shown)


peterpp:
Condition for non-GET request is not sufficient in general. For example `POST /_flush` with empty body also fails.

The real problem is the default value `$content = array()` together with null checking:
`$opts['http']['content'] = $content === null ? $content : json_encode($content);`

I created a merge request for Elasic long time ago with multiple fixes and this is my solution for 7.14 compatibility:
https://github.com/vrana/adminer/pull/420/commits/920a11c3954b9f3187bab5bc9aab7d6ee399879d

peterpp:
As long as Adminer is not maintained anymore, you can find all my Elasticsearch fixes here: https://github.com/pematon/adminer

tauh33dkhan:
Thanks @peterpp

vrana:
Thanks.

PR #438, opened 2021-08-31
Function to retrieve driver name

Roy-Orbison:
Plugins cannot access `$drivers` global after compilation. This would allow display of the nice driver name without having to know the shrunken variable name.


vrana:
Thanks.

PR #440, opened 2021-09-18
Elasticsearch 7 driver

peterpp:
In Elasticsearch 7, mapping types are deprecated and there is only one system '_doc' mapping type. It does not make sense to display it anymore. So the whole presentation of Elastic indices in Adminer can be much improved. Indices can be treated as SQL tables instead of databases.

As long as this new approach is completelly different, it would by complicated to handle it by conditions in the current Elastic driver. And I honestly don't want to bother with it. So I created a brand new driver `elastic7` as a plugin and use it happily.

Note that this branch is rebased onto `field-priviliges` (https://github.com/vrana/adminer/pull/422) which is required for this driver to work corretly.


standa:
@peterpp any chance you could make this PR against https://github.com/adminerevo/adminerevo so it could become a part of a regular release?

peterpp:
Yes, I noticed activity around adminerevo and I'm happy that somebody is willing to do at least basic maintenance. I am quite busy right now, but I will prepare it.

peterpp:
@standa My Elasticseach drivers are on the way to AdminerEvo. However, the process is slow. Feel free to check out my own fork of the original Adminer with bunch of other fixes 😉

PR #441, opened 2021-09-24
Fix warning `$sql` not exists

zzjin:
As using php8+, call to undefined var will cause warning output, which cause mysql innodb table not shown table info(shown `?` instead of real data).

```
Mysql: MySQL version: 8.0.26 through PHP extension PDO_MySQL
PHP: 8.0.11
```


MosiiAUT:
Will this be implemented? Same as #450

Roy-Orbison:
@LionelLaffineur This is an old bug with an easy fix. Could you merge this PR into your fork?

vrana:
Thanks.

PR #429, opened 2021-05-22
Avoid Warning of PHP 8 on unknown properties (bug #801)

FrancoisCapon:
### Avoid Warning of PHP 8 on unknown properties on the `$field` object in function `select()`

@hemroy21 has open an issue on my "adminer sqlite repository":

- FrancoisCapon/LoginToASqlite3DatabaseWithoutCredentialsWithAdminer/issues/3

and on SourceForge:

- [#801 Undefined property: stdClass::$orgtable warning in adminer 4.8.1 ](https://sourceforge.net/p/adminer/bugs-and-features/801/)

We made some tests and the conclusion is that the issue is not specific to `sqlite` but only to `PHP 8.0` (Warning vs Notice).

```php
<?php
highlight_file(__FILE__);
echo '<p>'.PHP_VERSION.'</p>';
$object = (object) array ();
var_dump($object);
var_dump($object->property);

7.4.15

object(stdClass)#1 (0) { }
Notice: Undefined property: stdClass::$property in ... on line 6
NULL

8.0.2

object(stdClass)#1 (0) { }
Warning: Undefined property: stdClass::$property in ... on line 6
NULL
```

-  **Therefore, each unknown property must be set with PHP 8.0 to avoid warnings.**

PR #442, opened 2021-09-27
Prevention of the warning caused by an access to an undefined property (php8)

pospjan:
Access to an undefined property was a _notice_ in php 7.4, but was changed to a _warning_ in php 8. And as the Adminer is "hiding" only notices, this warning has started to appear while submitting custom queries:

Screenshot 2021-09-27 at 20 20 12

### DIFF
--- a/adminer/include/editing.inc.php
+++ b/adminer/include/editing.inc.php
@@ -49,7 +49,7 @@ function select($result, $connection2 = null, $orgtables = array(), $limit = 0) 
 				if ($field->charsetnr == 63) { // 63 - binary
 					$blobs[$j] = true;
 				}
-				$types[$j] = $field->type;
+				$types[$j] = isset($field->type) ? $field->type : null;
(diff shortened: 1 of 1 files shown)


jstephens7:
I'm facing this issue as well. A good alternative I've used in the mean time is to run this in the console (CTRL+SHIFT+I) to remove those warnings in the table headers.

`var children = document.querySelectorAll('th');
for(var i = 0; i < children.length; i++){
    var newText = children[i].innerHTML.replace(`<br>
<b>Warning</b>:  Undefined property: stdClass::$type in <b>/var/www/html/adminer.php</b> on line <b>1484</b><br>`,'');
    children[i].innerHTML = newText;
}`

pospjan:
I found out that this problem is better (more universally) solved in https://github.com/vrana/adminer/pull/429

guilliamxavier:
So close this PR and let's get #429 merged please

AnnoyingTechnology:
@pospjan Issue doesn't seem to be fixed with SQLite and latest adminer (`4.8.1`) + PHP `8.2.12`. 

I can reproduce exactly what is stated here : https://sourceforge.net/p/adminer/bugs-and-features/801/ 

If someone can re-open.

PR #444, opened 2021-09-30
mysqli_real_connect: set connection flag

romain-dartigues:
In my use case I have to connect to some MySQL services only available through a SSL connection but without a way to get their CA to trust.

I did not find any other way to fix the error:

> Connections using insecure transport are prohibited while --require_secure_transport=ON.

Than changing the [`mysqli_real_connect`](https://www.php.net/mysqli.real-connect#refsect1-mysqli.real-connect-parameters) connection flags from `64` (`MYSQLI_CLIENT_SSL_DONT_VERIFY_SERVER_CERT`) to `2048` (`MYSQLI_CLIENT_SSL`), i.e.:

```php
<?php
function adminer_object() {
	include_once "./plugins/plugin.php";
	$adminer = new AdminerPlugin(Array());
	foreach (glob("plugins/*.php") as $filename) include_once "./$filename";
	$adminer->plugins = Array(
		new AdminerLoginSsl(Array("key"=>NULL, "cert"=>NULL, "ca"=>NULL, "cflags"=>MYSQLI_CLIENT_SSL)),
		// ...
	);
	return $adminer;
}
include "./adminer.php";
```

Any thoughts on this problem and other possibles workarounds?

**Warning:** this patch use the PHP 7 [Null Coalesce Operator](https://wiki.php.net/rfc/isset_ternary) syntax.

### DIFF
--- a/adminer/drivers/mysql.inc.php
+++ b/adminer/drivers/mysql.inc.php
@@ -27,7 +27,7 @@ if (!defined("DRIVER")) {
 					$database,
 					(is_numeric($port) ? $port : ini_get("mysqli.default_port")),
 					(!is_numeric($port) ? $port : $socket),
-					($ssl ? 64 : 0) // 64 - MYSQLI_CLIENT_SSL_DONT_VERIFY_SERVER_CERT (not available before PHP 5.6.16)
+					($ssl?$ssl['cflags'] ?? 64 : 0) // 64 - MYSQLI_CLIENT_SSL (not available before PHP 5.6.16)
 				);
 				$this->options(MYSQLI_OPT_LOCAL_INFILE, false);
 				return $return;


colinmollenhour:
I also cannot connect with SSL with Adminer even though I can connect with the mysql cli using the same certificate.. I haven't tried this but it would be nice to be able to set flags all the same.

romain-dartigues:
@vrana < do you want me to work on a better PR (and do you have guidelines for it) or is it a lost cause?

colinmollenhour:
Just FYI, not sure if this helps your case or not but the adminer docker container maintainer accepted a PR to replace PDO with mysqli as the default installed driver: https://github.com/TimWolla/docker-adminer/pull/109

romain-dartigues:
Thank you for the information @colinmollenhour.
Unfortunately it does not relate to case as we are not using an OCI image.

peterpp:
@romain-dartigues Do you still need it? I'am considering this configuration to my fork at https://github.com/pematon/adminer

eyupdzhanY1:
Have you added this change @peterpp  ?

peterpp:
@eyupdzhanY1 Not yet. I'm not sure if I fully understand your problem. You don't want to set custom certificate by `mysqli::ssl_set` but you still need to set `MYSQLI_CLIENT_SSL` flag explicitly? As I understand this flag is used for establishing SSL connection and [according to the documentation](https://www.php.net/manual/en/mysqli.constants.php#constant.mysqli-client-ssl), it is set automatically.

eyupdzhanY1:
@peterpp  I am trying to troubleshoot, my server does not have CA , but ssl is enabled on server side so i get error
Connections using insecure transport are prohibited while --require_secure_transport=ON.
@romain-dartigues suggests that it fixed his issues with the error message so i wanted to try.

romain-dartigues:
Hello there, I'm not really needing it (and at the time I was patching it on the fly thanks to `sed` power anyway) but I though other people would be in the same situation I was in (confer the original post).
Maybe @eyupdzhanY1 could confirm if this workaround is still working or if anything else is better.

vrana:
#483 seems better.

PR #446, opened 2021-11-10
When editing turn URLs into clickable links

domlat:
Maybe there's a reason this isn't being done but one of our users requested that the output of stored procedures should produce clickable links.


nitodico:
This is caused by the vertical align attribute to INPUT and not SELECTs

imagen

vrana:
This check is already a couple of lines above.

PR #447, opened 2021-11-17
support for array values in clickhouse driver

morozovsk:


pivolan:
does it column read only mode? Will it fails if somebody try to edit and save cell?

morozovsk:
UPDATE doesn't work in clickhouse.
https://stackoverflow.com/questions/37901642/updating-data-in-clickhouse

vrana:
Thanks.

PR #451, opened 2022-02-05
Can we make adminer connect to mongodb databases using database that inputed?

bijancot:
just wondering,

can we make adminer connect to mongodb databases using inputed database as auth instead of `admin` database?

In some case, users authenticated by database that have been defined. So when this condition appear, we cannot login to database until we add that users to `admin` database too﻿


vrana:
Thanks.

PR #452, opened 2022-02-10
MySQL PDO Add possibility to disable verification of the server SSL certificate.

araisch:
While using explicitly `mysqli` as driver the server certificate verification is switched to false by default, using Standard driver `PDO` it can't be disabled. There are some use-cases making that the only way to use TLS (e.g. running with Self-Signed Certificates in Docker).
No PHP-Skills, so please check :)

### DIFF
--- a/adminer/drivers/mysql.inc.php
+++ b/adminer/drivers/mysql.inc.php
@@ -246,6 +246,9 @@ if (!defined("DRIVER")) {
 					if (!empty($ssl['ca'])) {
 						$options[PDO::MYSQL_ATTR_SSL_CA] = $ssl['ca'];
 					}
+					if (!empty($ssl['verify'])) {
+						$options[PDO::MYSQL_ATTR_SSL_VERIFY_SERVER_CERT] = $ssl['verify'];
+					}
 				}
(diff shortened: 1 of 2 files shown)


Kabingeran:
Please add for MS SQL also. You should get something like: TrustServerCertificate=yes in connection

araisch:
@Kabingeran Since this seems pretty abandonded and I have no MS SQL to check and additionally I am no php-coder at all, I can't help you. Sorry.

araisch:
@Kabingeran To be more precisely: I made this PR because I was in need for that feature and wanted to preserve others from searching again. I have absolutely no capacity for doing anything **even if** I had the skills and the environment needed..

Kabingeran:
OK. I'll do it myself if anyone else votes for this functionality

gth:
Not sure if you're still willing to accept votes, but count me as +1 for Adminer adding the TrustServerCertificate=yes option for SQL Server.

Kabingeran:
@gth
https://github.com/vrana/adminer/pull/479

vrana:
Merged, thanks.

PR #453, opened 2022-03-18
Microsoft SQL Unicode String Handling

HyP3r-:
With this Pull Request I have fixed the unicode string handling. All unicode strings are prefixed with a "N" so they are treated correct.

### DIFF
--- a/adminer/drivers/mssql.inc.php
+++ b/adminer/drivers/mssql.inc.php
@@ -40,7 +40,8 @@ if (isset($_GET["mssql"])) {
 			}
 
 			function quote($string) {
-				return "'" . str_replace("'", "''", $string) . "'";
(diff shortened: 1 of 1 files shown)


its0x08 (approved):

peterpp:
@HyP3r- What about to quote everything with `N` notation without the condition? I don't use MS SQL but I would like to merge this patch into my fork.

HyP3r-:
@peterpp In principle, I don't think there's anything wrong with that. But I really can't say what side effects it might have. And since I didn't know whether it would have any side effects, I programmed/built it that way.

peterpp:
@HyP3r- Let's keep it as you proposed. Merged to https://github.com/pematon/adminer if you are interested.

vrana:
Merged, thanks.

PR #455, opened 2022-06-10
Update hydra theme to most recent version

patcon:
Thanks for maintaining this!

The current hydra theme version is 2 years old. There's a horizontal scroll bug that prevents view columns beyond edge of screen, esp difficult for small screens with big tables. This fixes that.

Thanks!


PilaScat:
why?

vrana:
Because Adminer doesn't depend on them.

PR #461, opened 2022-11-24
CockroachDB Support.

gnat:
Work in progress CockroachDB support.

Only needed a few changes to become mostly functional, based on the Postgres driver.


vrana:
This seems far from finished.

PR #464, opened 2022-12-29
Replace for PHP tmpfile function

unix4you2:
Some shared hostings doesn't allow to use the tmpfile() PHP functions (i.e. when this function is disabled via  disable_functions  in the php.ini), so Adminer stop working in that servers.  This change create a random name for a temporary file that use the standard fopen function.


unix4you2:
A better version will be sent with auto-delete temporary file.

PR #465, opened 2022-12-29
Replace for PHP tmpfile function

unix4you2:
Some shared hostings doesn't allow to use the tmpfile() PHP functions (i.e. when this function is disabled via disable_functions in the php.ini), so Adminer stop working in that servers. This change create a random name for a temporary file that use the standard fopen function.

### DIFF
--- a/adminer/include/tmpfile.inc.php
+++ b/adminer/include/tmpfile.inc.php
@@ -5,7 +5,15 @@ class TmpFile {
 	var $size;
 	
 	function __construct() {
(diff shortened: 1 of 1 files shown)


zidanpragata:
$zh=$b->importServerPath();
file_exists($zh)?$zh:"compress.zlib://$zh.gz"),"rb");

existing code is sophisticated, administrator doesn't save uploaded files in the trash. because it avoids the danger of root access due to admin carelessness.

PR #474, opened 2023-07-19
Updates pepa-linha submodule to include latest changes

SirTimme:
I tried to use the `pepa-linha-dark` theme for adminer and noticed some css issues (see image below)

image

After testing with the latest css-File from https://github.com/pepa-linha/Adminer-Design-Dark/blob/master/adminer.css the issue resolved, so the submodule tracks an old commit.

image

This PR updates the submodule to the latest commit.

Im not sure if i specified the path correctly in `.gitmodules` so feel free to correct me


Baltix:
Original project is no longer maintained, please send your commits to AdminerEvo - https://github.com/adminerevo/adminerevo

SirTimme:
Will do, thanks for info!

PR #475, opened 2023-07-31
PostgreSQL: only show databases you can access

thomas-daniels:
When connecting to a PostgreSQL database, Adminer's dropdown also includes all databases you don't have access to, which is not very useful. This commit filters out databases that you don't have the USAGE privilege on, to present a cleaner overview of available databases.


peterpp:
Merged to https://github.com/pematon/adminer/commit/0d0936550cdb9231167500a029f647409042b243 with few changes. I've kept `has_database_privilege` in condition and added `datallowconn` (to hide `template0` database).

vrana:
Thanks, merged.

PR #476, opened 2023-08-08
Woocommerce compatibility fix

codezz:
Woocommerce also adds the is_ajax function and it will run into a fatal error


vrana:
Adminer is not meant to be used together with other projects.

PR #483, opened 2023-12-02
Fix mysqli ssl without server certificate

zeleznypa:
When I need to connect to Mysql server via SSL but I don't have a server certificate, there is a possibility to setup the connection like the following:

```php
function adminer_object()
{
    // autoloader
    foreach (glob(__DIR__ . '/plugins/*.php') as $filename) {
        require_once $filename;
    }

    $plugins = [
        new AdminerLoginSsl(['key' => '', 'cert' => '', 'ca' => '']),
    ];

    return new AdminerPlugin($plugins);
}

require __DIR__ . '/adminer.php';
```

But unfortunately, it doesn't work with the `MYSQLI_CLIENT_SSL_DONT_VERIFY_SERVER_CERT` directive. It fails on the error: `Connections using insecure transport are prohibited while --require_secure_transport=ON.`

But with the `MYSQLI_CLIENT_SSL` it works like a charm.

### DIFF
--- a/adminer/drivers/mysql.inc.php
+++ b/adminer/drivers/mysql.inc.php
@@ -27,7 +27,7 @@ if (!defined("DRIVER")) {
 					$database,
 					(is_numeric($port) ? $port : ini_get("mysqli.default_port")),
 					(!is_numeric($port) ? $port : $socket),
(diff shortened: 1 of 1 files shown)


vrana:
Merged, thanks.

PR #484, opened 2023-12-17
PHP 8.3 error suppression

Sneda8:
PHP 8.3 has shortened the array access on null error message to "Trying to access array offset on null". This commit changes the regular expression used to circumvent errors.

### DIFF
--- a/adminer/include/bootstrap.inc.php
+++ b/adminer/include/bootstrap.inc.php
@@ -1,6 +1,6 @@
 <?php
 function adminer_errors($errno, $errstr) {
-	return !!preg_match('~^(Trying to access array offset on value of type null|Undefined array key)~', $errstr);
+	return !!preg_match('~^(Trying to access array offset on( value of type)? null|Undefined array key)~', $errstr);
 }
 
 error_reporting(6135); // errors and warnings


jeffwhansen:
+1

diam:
Thank you very much for this correction.
When will you be able to publish an updated version, please?

SonnySVK (approved):

sktnetwork (approved):

nbao:
please release a new version compatible with php 8.3

SkyEmie:
Just for personal curiosity, why the choice is to catch these errors in this way, instead of adapting the code to avoid generating theses errors?

peterpp:
Thanks for patch, merged to https://github.com/pematon/adminer

@SkyEmie Those warnings are suppressed to simplify the code, especially on PHP 5. And also to introduce a way, how to make unintended typing mistakes very easily :)

SkyEmie:
The repo doesn't seem to be maintained since 3 years
I suggest using the fork https://github.com/adminerevo/adminerevo instead

peterpp:
> The repo doesn't seem to be maintained since 3 years I suggest using the fork https://github.com/adminerevo/adminerevo instead

I know and I don't want to use AdminerEvo. That's why I'm patching my own fork... (feel free to check it 😉).

TimexPeachtree (approved):

ShaiMagal:
it's sad that some forks have to be solved and not just one... It would be better to solve it in one place (here - official place), it's a shame that Mr. Vrána (@vrana) doesn't have time to approve "new" PRs :(

vrana:
Merged, thanks.

PR #485, opened 2023-12-21
Sql command pos

cesarep:
Merely changes the SQL command textbox to above the results, instead of below.
It allows for easier resizing of the textbox and quicker edit of the query.


vrana:
I'm not interested in this.

PR #486, opened 2024-01-02
PostgreSQL: Make data length calculation more accurate

caltong:
According to my testing and PG doc below, the original data length **does not** include the size of TOAST, while the index length, on the contrary, includes the size of TOAST.

In this PR, I use the pg_table_size and pg_indexes_size to make data length and index length more accurate.

> The TOAST management code is triggered only when a row value to be stored in a table is wider than TOAST_TUPLE_THRESHOLD bytes (normally 2 kB). The TOAST code will compress and/or move field values out-of-line until the row value is shorter than TOAST_TUPLE_TARGET bytes (also normally 2 kB, adjustable) or no more gains can be had. During an UPDATE operation, values of unchanged fields are normally preserved as-is; so an UPDATE of a row with out-of-line values incurs no TOAST costs if none of the out-of-line values change.

> pg_table_size ( regclass ) → bigint. Computes the disk space used by the specified table, excluding indexes (but including its TOAST table if any, free space map, and visibility map).

### DIFF
--- a/adminer/drivers/pgsql.inc.php
+++ b/adminer/drivers/pgsql.inc.php
@@ -322,7 +322,7 @@ ORDER BY 1";
 
 	function table_status($name = "") {
 		$return = array();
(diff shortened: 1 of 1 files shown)


peterpp:
Looks good, merged to https://github.com/pematon/adminer

vrana:
Merged, thanks.

PR #490, opened 2024-06-02
Fixed string constants and queries splitting.

alxivnov:
`\` at the end of string constants breaks string/query splitting. Here is the test:
```sql
-- \ at the end of strting query splitting
SELECT 'some''', '\', ';';

SELECT E'some\'', E'\\', E';';

SELECT 'some''', '\', ';', E'some\'', E'\\', E';';

SELECT E'some\'', E'\\', E';', 'some''', '\', ';';
```

### DIFF
--- a/DOCKER.md
+++ b/DOCKER.md
@@ -1,0 +1,31 @@
+```shell
+# Remove previous container
+docker rm -f adminer-php-9999
+
+# Start PHP 5.6 container
+docker run \
+	--log-driver local \
+	--name adminer-php-9999 \
+	--publish=9999:8000 \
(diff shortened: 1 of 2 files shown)


peterpp:
Thanks for sharing, but your fix contains several problems:
* C-style escapes depends on SQL dialect and server settings.
* String constants letter `E` is PostgreSQL specific and is case-insensitive.
* Added `"[^']"` rule works only for single quotes, but main cycle detects various types of string quotes.

Here you can find a solution based on your work (if you are interested): https://github.com/pematon/adminer/commit/707de6000c9fda5d0a07e52e8f88ad0727feb7e3

vrana:
Thanks.

PR #491, opened 2024-09-23
MongoDB update not working

herobank110:
'WHERE' parsing was removing the closing paren of object IDs

eg `_id = MongoDB\BSON\ObjectID("66d5e4468b906497efb4cb93"`


peterpp:
Thanks for sharing. I'm not sure why that regexp removes parenthesis. I suppose that the condition can be `"WHERE ((a=1) OR (b=2))"`, so I've kept previous logic in my fix here: https://github.com/pematon/adminer/commit/63c0581630a0e2bc792d17403318b8d7be615d71

vrana:
Merged, thanks.

PR #492, opened 2024-11-02
feat: uuid for MariaDB >= 10.7

vukbgit:
added for MariaDB >= 10.7 (see [data type doc](https://mariadb.com/kb/en/uuid-data-type/)):

- uuid string data type
- uuid() function for uuid type on new/edit item form

### DIFF
--- a/adminer/drivers/mysql.inc.php
+++ b/adminer/drivers/mysql.inc.php
@@ -366,7 +366,7 @@ if (!defined("DRIVER")) {
 	* @return mixed Min_DB or string for error
 	*/
 	function connect() {
-		global $adminer, $types, $structured_types;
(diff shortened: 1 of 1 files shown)


peterpp:
Thank you for sharing, I merge it into our fork here: https://github.com/pematon/adminer, in case you are interested.

I also:
- fixed version checking condition because `min_version(0, 10.7, $connection)` would allow UUID data type also for MySQL
- change uuid() default value to a function selected in functions list while adding new row.

vukbgit:
thanks, keep up the great work! I'm gonna try https://github.com/pematon/adminer v5 asap

vrana:
Merged, thanks.

PR #496, opened 2025-02-27
Add Reset button for tables-filter plugin

Denitz:
Add simple reset button displayed on none-empty input, modernize JS a bit:

image


vrana (changes requested):
With `type="search"`, the cross should be rendered by the browser. This is quite ugly.

vrana:
Merged, thanks.

PR #497, opened 2025-02-27
New version of design rmSOFT

mesaros:
New version of design rmSOFT


vrana:
Merged, thanks. I've kept the designs in their current directories because they are linked from elsewhere.
I've also put each rule on a separate line to make this diffable.

PR #498, opened 2025-02-28
Elastic: Properly display sparse result rows

cweiske:
Result records in Elasticsearch do not always have all columns that are defined in an index.
This often happens when multiple document types are stored in the same index.

The first row has columns ["_id", "html", "url"], while the second misses the "html" column: ["_id", "url"].

Up to now, Adminer expected that all result rows include all columns. This led to the problem that the "url" value in the 2nd example row was rendered in the "html" column.

This patch fixes this problem by iterating over the expected keys, and filling NULL when the database result row misses this key.

(I tried to fix that in the Elasticsearch driver only, but this is much more complicated, since MinDriver::select() does not know about expected field names at all.)

I've verified that MySQL listings look the same as before.

# Screenshots
## Manual column selection: all fine
2025-02-28 adminer elastic list - manual filtering

## Automatic column selection: bug
Values are rendered in the wrong column
2025-02-28 adminer elastic list - standard view

### DIFF
--- a/plugins/drivers/elastic.php
+++ b/plugins/drivers/elastic.php
@@ -100,7 +100,21 @@ if (isset($_GET["elastic"])) {
 
 			function __construct($rows) {
 				$this->num_rows = count($rows);
-				$this->_rows = $rows;
+
+				$keys = array();
+				foreach ($rows as $row) {
+					foreach ($row as $key => $val) {
+						$keys[$key] = true;
+					}
+				}
+				$this->_rows = array();
+				foreach ($rows as $row) {
+					$insert = array();
+					foreach ($keys as $key => $val) {
+						$insert[$key] = $row[$key];
+					}
+					$this->_rows[] = $insert;
+				}
 
 				reset($this->_rows);
 			}


adminer/select.inc.php
```diff
@@ -388,7 +388,8 @@
 					. ($is_group || information_schema(DB) ? "" : " <a href='" . h(ME . "edit=" . urlencode($TABLE) . $unique_idf) . "' class='edit'>" . lang('edit') . "</a>")
 				);
 
-				foreach ($row as $key => $val) {
+				foreach ($names as $key => $dummy) {
+					$val = $row[$key] ?? null;
```
vrana, line 392: This isn't supported by PHP 5 and it's also useless.

adminer/select.inc.php
```diff
@@ -388,7 +388,8 @@
...
 					if (isset($names[$key])) {
```
vrana, line 393: This is now never false.

vrana (changes requested):
`$names` is populated from the first row. If it doesn't contain some column then this column is not printed for any row.

vrana:
I think that this code added to `Result::__construct` should do the trick:

```php
foreach ($rows as $row) {
	foreach ($row as $key => $val) {
		$keys[$key] = true;
	}
}
foreach ($rows as $row) {
	$insert = array();
	foreach ($keys as $key => $val) {
		$insert[$key] = $row[$key];
	}
	$this->_rows[] = $insert;
}
```

cweiske:
Yes, this would work. The downside is that the original order of the fields in database is not preserved, which I wanted to prevent.

But since `$names` - as you correctly noted - are generated from the first row, the original field order is not available anywhere. Thus we have to accept that minor problem to fix the big issue.

cweiske:
Hm. But I could also fetch the fields in `select()`.

cweiske:
I think that #893 is a better solution.

PR #501, opened 2025-03-03
Fix connecting to a MySQL server that has a self-signed SSL certificate

joshbmarshall:
Both the MySQLi and PDO_MYSQL interfaces had minor issues that did not allow the use of a self-signed SSL certificate on the database server.

For PDO interface the plugin config needs to set the verify to false. For MySQLi the config needs to not specify a certificate.

### DIFF
--- a/adminer/drivers/mysql.inc.php
+++ b/adminer/drivers/mysql.inc.php
@@ -246,7 +246,7 @@ if (!defined("DRIVER")) {
 					if (!empty($ssl['ca'])) {
 						$options[PDO::MYSQL_ATTR_SSL_CA] = $ssl['ca'];
 					}
-					if (!empty($ssl['verify'])) {
+					if (isset($ssl['verify'])) {
 						$options[PDO::MYSQL_ATTR_SSL_VERIFY_SERVER_CERT] = $ssl['verify'];
 					}
 				}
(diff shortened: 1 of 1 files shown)


adminer/drivers/mysql.inc.php
```diff
@@ -27,7 +27,7 @@ function connect($server = "", $username = "", $password = "", $database = null,
 					$database,
 					(is_numeric($port) ? $port : ini_get("mysqli.default_port")),
 					(!is_numeric($port) ? $port : $socket),
-					($ssl ? (empty($ssl['cert']) ? 2048 : 64) : 0) // 2048 - MYSQLI_CLIENT_SSL, 64 - MYSQLI_CLIENT_SSL_DONT_VERIFY_SERVER_CERT (not available before PHP 5.6.16)
+					($ssl ? (!empty($ssl['cert']) ? 2048 : 64) : 0) // 2048 - MYSQLI_CLIENT_SSL, 64 - MYSQLI_CLIENT_SSL_DONT_VERIFY_SERVER_CERT (not available before PHP 5.6.16)
```
vrana, line 30: This one should use `$ssl['verify']`.

vrana:
Fixed in other way.

joshbmarshall:
You missed the PDO fix in your other way of fixing.

Also what if $ssl['verify'] is not defined? As it's a new variable it may break backwards

PR #891, opened 2025-03-10
Make login to Elasticsearch work again

cweiske:
By partially reverting the changes made by commit 2c9f380c64ec84a1dc44745df29c5fbf1081fcb0
> Elastic: Remove plugin for version 5

which removed the passwordless login plugin from `adminer/elastic.php`.

Since the passwordless plugin is still enabled for Sqlite, I think the removal for elastic was a mistake.

### DIFF
--- a/adminer/elastic.php
+++ b/adminer/elastic.php
@@ -2,8 +2,13 @@
 // To create Adminer just for Elasticsearch, run `../compile.php elastic`.
 
 function adminer_object() {
+	include_once "../plugins/plugin.php";
+	include_once "../plugins/login-password-less.php";
 	include_once "../plugins/drivers/elastic.php";
-	return new Adminer\Adminer;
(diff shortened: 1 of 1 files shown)


vrana:
Elasticsearch has [security mode](https://www.elastic.co/guide/en/elasticsearch/reference/current/security-minimal-setup.html), enabled by default since version 8. I think that the proper solution would be to have an optional password.

cweiske:
I don't really understand what you mean by "optional password".

To support both ES7 (no password) and ES8 (with password), both the es7-passwordless-password and the es8-real-password need to be handled. But this is done already in the passwordless plugin, which returns the password as part of the credentials when it's not the passwordless-password.

Other than that I don't know how it should work since I'm not aware that adminer has a configuration file.

vrana:
You are right.

PR #900, opened 2025-03-10
Compile: supports visibility modifiers

dg:
Unfortunately, I can't test it because for some reason the compiler is generating invalid files:

```php
<?php
/** Adminer - Compact database management
* @link https://www.adminer.org/
* @author Jakub Vrana, https://www.vrana.cz/
* @copyright 2007 Jakub Vrana
* @license https://www.apache.org/licenses/LICENSE-2.0 Apache License, Version 2.0
* @license https://www.gnu.org/licenses/gpl-2.0.html GNU General Public License, version 2 (one or other)
* @version 5.0.2
*/namespace
Adminer;namespace
Adminer;namespace
Adminer;$ia="5.0.2";namespace
```

### DIFF
--- a/compile.php
+++ b/compile.php
@@ -297,7 +297,7 @@ function php_shrink($input) {
 				$doc_comment = true;
 				$token[1] = substr_replace($token[1], "* @version $VERSION\n", -2, 0);
 			}
-			if ($token[0] == T_VAR) {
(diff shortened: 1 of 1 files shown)


vrana:
Thank you. The doubled namespace could be because of `\r\n` used at line ends.

PR #919, opened 2025-03-17
Function to retrieve driver name

Roy-Orbison:
This was already merged in PR #438, but broken in fe88f83. Since API has had breaking changes for v5.x, may as well have a better function name.


vrana:
I'm sorry about that, I'll return the function with original name.

Roy-Orbison:
Thanks, good call.

PR #936, opened 2025-03-21
Spanish translation upgrade

unix4you2:
Just some edditions and new translations


vrana:
Thanks, there are some other new messages. Run `lang.php es` to get them. There's also a missing singular for “Muchos intentos de acceso Intente en %d minutos.”

PR #954, opened 2025-03-25
fix: icons background in lucas-sandery design

vukbgit:
hello, with lucas-sandery theme icons for adding/moving/removing fields are not displayed into "alter table" page . I fixed the problem by commenting height and padding-top declarations for .icon class .


vukbgit:
ok, checked your more elegant solution (I'm not into theme design, at least not yet), pulled theme again, works like a charm, thank you for the fix and for the Fantastic Adminer, cheers

PR #961, opened 2025-03-29
New version of design rmSOFT (5.1.0.1)

mesaros:
New version of design rmSOFT (5.1.0.1)


vrana:
Merged, thanks. Please also add `--bg` and `--fg` variables to be forwards compatible. https://github.com/vrana/adminer/blob/master/designs/rmsoft_blue-dark/adminer-dark.css#L39

PR #970, opened 2025-04-02
Use values from select to populate order_columns with aggregated columns

hjsleijster:
It adds the aggregated columns to the sort selectboxes.
This makes it possible to sort on multiple columns, where at least one of them is aggregated.


hjsleijster:
Thanks for adding this in 6.0.0!

PR #974, opened 2025-04-03
New version of design rmSOFT

mesaros:
New version of design rmSOFT


vrana:
There are some `map-get` in the styles which is SASS specific.

Also, can you please always pull request only one commit, not the whole merge branch?

vrana:
Merged with fixed `--bg` and `--fg`, thanks.

PR #984, opened 2025-04-04
Wrap Loaded plugins into div

RiKap:
Sometimes it is not that important to have this info.
It will just make easier to move it, hide it, with css instead of js hacks like [this one](https://github.com/dg/adminer/commit/980f57661e170a0ab48bbe1ac34026e4ec5768af).


wintstar:
Line 48 Tag `<p>` is not closed `</p>`.

Depending on the number of plug-ins loaded, the list could become very long. It would be better if you could expand the list.
Ideal would be:
```
 echo "<details>\n";
    echo " <summary>" . lang('Loaded plugins') . "</summary>\n";
  	echo "<div class='plugins'>\n";
	echo "<ul>\n";
	foreach (adminer()->plugins as $plugin) {
		$reflection = new \ReflectionObject($plugin);
		echo "<li><b>" . get_class($plugin) . "</b>" . h(preg_match('~^/[\s*]+(.+)~', $reflection->getDocComment(), $match) ? ": $match[1]" : "") . "\n";
	}
	echo "</ul>\n";
	echo "</div>\n"
 echo "</details>\n";
```

vrana:
I've moved the plugins down in 48f82b34.

PR #990, opened 2025-04-06
Update Polish translation

Matthaiks:


vrana:
Thank you for your always prompt translations. If you wish, you can also translate plugins, e.g. [sql-gemini](https://github.com/vrana/adminer/blob/master/plugins/sql-gemini.php#L95), [config](https://github.com/vrana/adminer/blob/master/plugins/config.php#L70) or [menu-links](https://github.com/vrana/adminer/blob/master/plugins/menu-links.php#L70).

PR #991, opened 2025-04-06
Add Polish translation of plugins

Matthaiks:


vrana:
Thanks. [lang.php](https://github.com/vrana/adminer/blob/master/lang.php) now updates plugins too.

vrana:
You can also help with the web translations if you wish: https://www.adminer.org/pl/translations/

PR #469, opened 2023-03-18
Add infrastructure for chosing index methods initial support for postgresql

salacr:
Most DB engines (at least Postgresql, MySQL, ORacle . SQLServer) support different index methods.
For Oracle, and SQLServer the syntax is more complicated but it's out of the scope infrastructure should be the same.

PR #1030, opened 2025-04-13
Add infrastructure for chosing index methods initial support for postgresql

salacr:
This is a new version of https://github.com/vrana/adminer/pull/469 which i closed by mistake during rebase.
Code style should be fixed

### DIFF
--- a/adminer/drivers/pgsql.inc.php
+++ b/adminer/drivers/pgsql.inc.php
@@ -338,6 +338,19 @@ if (isset($_GET["pgsql"])) {
 			return get_vals("SELECT relname FROM pg_inherits JOIN pg_class ON inhrelid = oid WHERE inhparent = " . $this->tableOid($table) . " ORDER BY 1");
 		}
 
+		function partitionsInfo(string $table): array {
+			$row = connection()->query("SELECT * FROM pg_partitioned_table WHERE partrelid = " . driver()->tableOid($table))->fetch_assoc();
+			if ($row) {
+				$attrs = get_vals("SELECT attname FROM pg_attribute WHERE attrelid = $row[partrelid] AND attnum IN (" . str_replace(" ", ", ", $row["partattrs"]) . ")"); //! ordering
+				$by = array('h' => 'HASH', 'l' => 'LIST', 'r' => 'RANGE');
+				return array(
+					"partition_by" => $by[$row["partstrat"]],
+					"partition" => implode(", ", array_map('Adminer\idf_escape', $attrs)),
+				);
+			}
+			return array();
+		}
+
 		function tableOid(string $table): string {
 			return "(SELECT oid FROM pg_class WHERE relnamespace = $this->nsOid AND relname = " . q($table) . " AND relkind IN ('r', 'm', 'v', 'f', 'p'))";
 		}
@@ -943,12 +956,9 @@ AND typelem = 0"
 		}
 		$return .= implode(",\n    ", $return_parts) . "\n)";
 
-		$table_oid = driver()->tableOid($status['Name']);
-		$row = connection()->query("SELECT * FROM pg_partitioned_table WHERE partrelid = $table_oid")->fetch_assoc();
-		if ($row) {
-			$attrs = get_vals("SELECT attname FROM pg_attribute WHERE attrelid = $row[partrelid] AND attnum IN (" . str_replace(" ", ", ", $row["partattrs"]) . ")"); //! ordering
-			$by = array('h' => 'HASH', 'l' => 'LIST', 'r' => 'RANGE');
-			$return .= "\nPARTITION BY $by[partstrat](" . implode(", ", array_map('Adminer\idf_escape', $attrs)) . ")";
+		$partition = driver()->partitionsInfo($status['Name']);
+		if ($partition) {
+			$return .= "\nPARTITION BY $partition[partition_by]($partition[partition])";
 		}
 		//! parse pg_class.relpartbound to create PARTITION OF
 
--- a/adminer/lang/cs.inc.php
+++ b/adminer/lang/cs.inc.php
@@ -195,7 +195,6 @@ Lang::$translations = array(
 	'Partition name' => 'Název oddílu',
 	'Values' => 'Hodnoty',
 	'Inherits from' => 'Zděděná z',
-	'Inherited tables' => 'Zděděné tabulky',
 
 	'View' => 'Pohled',
 	'Materialized view' => 'Materializovaný pohled',
--- a/adminer/lang/pl.inc.php
+++ b/adminer/lang/pl.inc.php
@@ -196,7 +196,6 @@ Lang::$translations = array(
 	'Partition name' => 'Nazwa partycji',
 	'Values' => 'Wartości',
 	'Inherits from' => 'Dziedziczy po',
-	'Inherited tables' => 'Tabele dziedziczone',
 
 	'View' => 'Perspektywa',
 	'Materialized view' => 'Zmaterializowana perspektywa',
--- a/adminer/lang/xx.inc.php
+++ b/adminer/lang/xx.inc.php
@@ -197,7 +197,6 @@ Lang::$translations = array(
 	'Partition name' => 'Xx',
 	'Values' => 'Xx',
 	'Inherits from' => 'Xx',
-	'Inherited tables' => 'Xx',
 
 	'View' => 'Xx',
 	'Materialized view' => 'Xx',
--- a/adminer/table.inc.php
+++ b/adminer/table.inc.php
@@ -110,6 +110,10 @@ if (support(is_view($table_status) ? "view_trigger" : "trigger")) {
 
 $inherited = driver()->inheritedTables($TABLE);
 if ($inherited) {
-	echo "<h3 id='inherited'>" . lang('Inherited tables') . "</h3>\n";
+	echo "<h3 id='partitions'>" . lang('Partitions') . "</h3>\n";
+	$partition = driver()->partitionsInfo($TABLE);
+	if ($partition) {
+		echo "<p><code class='jush-" . JUSH . "'>BY " . h("$partition[partition_by]($partition[partition])") . "</code>\n";
+	}
 	tables_links($inherited);
 }


adminer/drivers/pgsql.inc.php
```diff
@@ -513,6 +514,32 @@ function indexes($table, $connection2 = null) {
...
+		static $methods = [];
+		if (!$methods) {
+
+			// default guess for old PG instances
+			$methods = array(
+				"btree",
+				"gin",
+				"gist"
+			);
+			if (min_version(9.6)) {
```
vrana, line 532: Why? I [can see](https://www.postgresql.org/docs/7.3/catalog-pg-am.html) this table in older versions too.

adminer/drivers/pgsql.inc.php
```diff
@@ -513,6 +514,32 @@ function indexes($table, $connection2 = null) {
 		return $return;
 	}
 
+	/**
+	 * return list of supported index methods first one is default
+	 * @return string[]
+	 */
+	function index_methods() : array
+	{
+		static $methods = [];
```
vrana, line 523: Define arrays with `array()`.

adminer/include/adminer.inc.php
```diff
@@ -355,7 +355,7 @@ function tableIndexesPrint(array $indexes): void {
 					. ($index["descs"][$key] ? " DESC" : "")
 				;
 			}
-			echo "<tr title='" . h($name) . "'><th>$index[type]<td>" . implode(", ", $print) . "\n";
+			echo "<tr title='" . h($name) . "'><th>$index[type]" . (index_methods() ? " (" . $index['method'] . ")" : "") . "<td>" . implode(", ", $print) . "\n";
```
vrana, line 358: This doesn't seem important enough to show here.

adminer/indexes.inc.php
```diff
@@ -29,6 +29,7 @@
 			$columns = array();
 			$lengths = array();
 			$descs = array();
+			$indexMethod = array_key_exists("method", $index) && in_array($index["method"], index_methods()) ? $index["method"] : "";
```
vrana, line 32: Use snake_case: `index_method`.

adminer/indexes.inc.php
```diff
@@ -52,14 +53,15 @@
 					&& array_values($existing["columns"]) === $columns
 					&& (!$existing["lengths"] || array_values($existing["lengths"]) === $lengths)
 					&& array_values($existing["descs"]) === $descs
-				) {
+                    && (array_key_exists("method", $existing) && $existing["method"] === $indexMethod)
```
vrana, line 56: Wrong indent.

adminer/indexes.inc.php
```diff
@@ -105,6 +107,11 @@
 <table class="nowrap">
 <thead><tr>
 <th id="label-type"><?php echo lang('Index Type'); ?>
+<?php
+if (index_methods()) {
+    echo "<th id=\"label-method\">" . lang('Index method');
```
vrana, line 112: Use `'` instead of `\"`.

adminer/indexes.inc.php
```diff
@@ -128,7 +135,11 @@
 	if (!$_POST["drop_col"] || $j != key($_POST["drop_col"])) {
 		echo "<tr><td>" . html_select("indexes[$j][type]", array(-1 => "") + $index_types, $index["type"], ($j == count($row["indexes"]) ? "indexesAddRow.call(this);" : ""), "label-type");
 
-		echo "<td>";
+        if (index_methods()) {
```
vrana, line 138: Make it visible only with Options checked.

adminer/indexes.inc.php
```diff
@@ -128,7 +135,11 @@
...
+            echo "<td>" . html_select("indexes[$j][method]", array_merge([""], index_methods()), $index['method'], $onchange = true, "label-method");
```
vrana, line 139: Delete `$onchange =`.

adminer/indexes.inc.php
vrana, line 112: Add message to `lang/xx.inc.php` to the right place.

vrana (changes requested):
Do you personally use this feature? I'm willing to include it if it has real users.

adminer/indexes.inc.php
salacr: Fixed, I wasn't sure if I should reuse idxopts class or create a new one and modify the JS function, so please what's your preference?

adminer/drivers/pgsql.inc.php
salacr: You are right the strucute is diferent in the past it contained only index access methods now it contains also table access methods but I improved it so it's compatible with both..

adminer/include/adminer.inc.php
salacr: TBH i think that the visibility is actually more important than the possibility to edit the index via adminer.
What about if we show only non-default values?
image

In most cases it wouldn't bother an one but if there will be some special type of index it will be visible at first glance

salacr:
> Do you personally use this feature? I'm willing to include it if it has real users.
Yes, it's quite handy GIN / GIST indexes are really common when working with PostGIS / geometry
HASH indexes are quite useful when I want to index tables for example with tokens

adminer/indexes.inc.php
vrana: Reuse.

adminer/include/adminer.inc.php
vrana: Yes, that's a good idea. I use the same concept e.g. for column collations.

adminer/drivers/pgsql.inc.php
```diff
@@ -324,6 +324,26 @@ function tableHelp(string $name, bool $is_view = false) {
...
+		/**
+		 * return list of supported index methods first one is default
+		 * @return string[]
+		 */
+		function indexMethods(): array {
+			static $methods = array();
+			if (!$methods) {
+				$amtype_condition = "";
+				if (min_version(9.6)) {
+					$amtype_condition = "WHERE amtype = 'i'";
```
vrana, line 336: Use `" WHERE..."` and remove the space from below.

adminer/drivers/pgsql.inc.php
```diff
@@ -324,6 +324,26 @@ function tableHelp(string $name, bool $is_view = false) {
...
+				}
+
+				$methods = array();
+				foreach (get_rows("SELECT amname FROM pg_am " . $amtype_condition . " ORDER BY amname='btree' DESC, amname ASC") as $row) {
```
vrana, line 340: Space around `=`, remove ` ASC`.

adminer/drivers/pgsql.inc.php
```diff
@@ -324,6 +324,26 @@ function tableHelp(string $name, bool $is_view = false) {
...
+		 */
+		function indexMethods(): array {
+			static $methods = array();
+			if (!$methods) {
+				$amtype_condition = "";
+				if (min_version(9.6)) {
+					$amtype_condition = "WHERE amtype = 'i'";
+				}
+
+				$methods = array();
```
vrana, line 339: Use `get_vals`.

adminer/lang/cs.inc.php
```diff
@@ -354,6 +354,7 @@
 	'%s must <a%s>return an array</a>.' => '%s musí <a%s>vracet pole</a>.',
 	'<a%s>Configure</a> %s in %s.' => '<a%s>Nakonfigurujte</a> %s v %s.',
 	'screenshot' => 'obrázek',
+	'Index Method' => 'Typ Indexu',
```
vrana, line 357: Small `m` in English, small `i` in Czech. I wanted you to update [xx.inc.php](https://github.com/vrana/adminer/blob/master/adminer/lang/xx.inc.php), you don't need to learn Czech 😉.

adminer/drivers/pgsql.inc.php
```diff
@@ -656,7 +677,7 @@ function alter_indexes($table, $alter) {
 			} elseif ($val[2] == "DROP") {
 				$drop[] = idf_escape($val[1]);
 			} else {
-				$queries[] = "CREATE INDEX " . idf_escape($val[1] != "" ? $val[1] : uniqid($table . "_")) . " ON " . table($table) . " (" . implode(", ", $val[2]) . ")";
+				$queries[] = "CREATE INDEX " . idf_escape($val[1] != "" ? $val[1] : uniqid($table . "_")) . " ON " . table($table) . ($val[3] ? " USING " . $val[3] : "") . " (" . implode(", ", $val[2]) . ")";
```
vrana, line 680: Update https://github.com/vrana/adminer/blob/v5.2.1/adminer/drivers/mysql.inc.php#L717 and make sure that PHPstan doesn't complain.

vrana, line 680: `" USING $val[3]"`

adminer/include/adminer.inc.php
```diff
@@ -346,6 +346,12 @@ function tableStructurePrint(array $fields, ?array $tableStatus = null): void {
 	*/
 	function tableIndexesPrint(array $indexes): void {
 		echo "<table>\n";
+		$index_methods = driver()->indexMethods();
+		$default_index_method = "";
+		if ($index_methods)
```
vrana, line 351: No newline before `{`.

adminer/include/adminer.inc.php
```diff
@@ -346,6 +346,12 @@ function tableStructurePrint(array $fields, ?array $tableStatus = null): void {
 	*/
 	function tableIndexesPrint(array $indexes): void {
 		echo "<table>\n";
+		$index_methods = driver()->indexMethods();
+		$default_index_method = "";
```
vrana, line 350: Just use `reset()` unconditionally.

adminer/include/adminer.inc.php
```diff
@@ -355,7 +361,10 @@ function tableIndexesPrint(array $indexes): void {
 					. ($index["descs"][$key] ? " DESC" : "")
 				;
 			}
-			echo "<tr title='" . h($name) . "'><th>$index[type]<td>" . implode(", ", $print) . "\n";
+
+			$index_method = ($index_methods && ($default_index_method != $index['method'])) ? " (" . $index['method'] . ")" : "";
```
vrana, line 365: `" ($index[method])"`

adminer/include/adminer.inc.php
```diff
@@ -355,7 +361,10 @@ function tableIndexesPrint(array $indexes): void {
...
+			echo "<tr title='" . h($name) . "'><th>$index[type]" . $index_method;
```
vrana, line 366: Put `$index_method` inside the string.

vrana, line 366: Put `<th>` in a separate `echo`.

adminer/drivers/pgsql.inc.php
```diff
@@ -324,6 +324,26 @@ function tableHelp(string $name, bool $is_view = false) {
 			}
 		}
 
+		/**
```
vrana, line 327: Delete duplicate doc-comment.

adminer/include/driver.inc.php
```diff
@@ -236,6 +236,14 @@ function supportsIndex(array $table_status): bool {
 		return !is_view($table_status);
 	}
 
+	/**
+	 * Return list of supported index methods, first one is default
+	 * @return string[]
```
vrana, line 241: `list<string>`

adminer/indexes.inc.php
```diff
@@ -29,6 +29,7 @@
 			$columns = array();
 			$lengths = array();
 			$descs = array();
+			$indexMethod = array_key_exists("method", $index) && in_array($index["method"], driver()->indexMethods()) ? $index["method"] : "";
```
vrana, line 32: Snake case with variables.

vrana, line 32: You don't need the `array_key_exists`.

adminer/indexes.inc.php
```diff
@@ -52,14 +53,15 @@
 					&& array_values($existing["columns"]) === $columns
 					&& (!$existing["lengths"] || array_values($existing["lengths"]) === $lengths)
 					&& array_values($existing["descs"]) === $descs
+					&& (array_key_exists("method", $existing) && $existing["method"] === $indexMethod)
```
vrana, line 56: You don't need `array_key_exists` if you initialize `$index_method` with `null`.

adminer/indexes.inc.php
```diff
@@ -105,6 +107,11 @@
 <table class="nowrap">
 <thead><tr>
 <th id="label-type"><?php echo lang('Index Type'); ?>
+<?php
+if (driver()->indexMethods()) {
+	echo "<th id='label-method' class='idxopts " .  ($show_options ? "" : " hidden") . "'>" . lang('Index Method');
```
vrana, line 112: We are on the Index page so `lang('Method')` seems sufficient here.

vrana, line 112: There's an extra space after `idxopts`.

adminer/indexes.inc.php
```diff
@@ -128,6 +135,10 @@
 	if (!$_POST["drop_col"] || $j != key($_POST["drop_col"])) {
 		echo "<tr><td>" . html_select("indexes[$j][type]", array(-1 => "") + $index_types, $index["type"], ($j == count($row["indexes"]) ? "indexesAddRow.call(this);" : ""), "label-type");
 
+		if (driver()->indexMethods()) {
+			echo "<td class='idxopts " .  ($show_options ? "" : " hidden") . "'>" . html_select("indexes[$j][method]", array_merge(array(""), driver()->indexMethods()), $index['method'], "label-method");
```
vrana, line 139: There's an extra space.

vrana:
> I was also thinking about adding support for partial indexes as well.

If it's not very intrusive to the UI.

adminer/drivers/pgsql.inc.php
salacr: I will fix that. I also wonder,  can't we utilize GitHub actions for PHPStan / PHPCode style? Quick POC looks promissing https://github.com/salacr/adminer/actions/runs/14461553951/job/40554974736

PR #1042, opened 2025-04-15
Add infrastructure for chosing index methods initial support for postgresql

salacr:
Hi, Sorry for new PR the previous one was closed?? When I sync my fork with your upstream (I must find out what's going on I'm quite new in GitHub workflow TBH)...
The previous PR is here https://github.com/vrana/adminer/pull/1030 I believe that I resolved all your comments.
Instead of Index Method I chose different wording (Algorithm), which is also used in SQL and is probably better.

Also at previous review I proposed usage of github actions to do PHPStan / PHPCS checks.
POC is here: https://github.com/salacr/adminer/actions/runs/14461710565

### DIFF
--- a/adminer/drivers/mssql.inc.php
+++ b/adminer/drivers/mssql.inc.php
@@ -409,6 +409,15 @@ WHERE OBJECT_NAME(i.object_id) = " . q($table), $connection2) as $row
 		return $return;
 	}
 
+	/**
+	 * return list of supported index methods first one is default
+	 * @return string[]
+	 */
+	function index_methods() : array
+	{
+		return array();
+	}
+
 	function view($name) {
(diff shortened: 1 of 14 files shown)


vrana:
Thank you for your work. There are still some nits but I'll fix them myself.

vrana:
This feels really nice, it's a good addition.

vrana:
Please also check if the index algorithm is exported.

salacr:
Yes, exports work as the index definition here is obtained via:

```
		foreach (get_rows("SELECT indexdef FROM pg_catalog.pg_indexes WHERE schemaname = current_schema() AND tablename = " . q($table) . ($primary ? " AND indexname != " . q($primary) : ""), null, "-- ") as $row) {
			$return .= "\n\n$row[indexdef];";
		}
```		
and indexdef contains all definition (even partial indexes will be working)

PR #1009, opened 2025-04-09
Allow designs to specify light and dark in one file

lucas-sandery:
Designs named `adminer-auto.css` would not have to provide both files. This enables simple rule precedence like so:

1. Light rule
2. Dark rule in media query
3. Rule that overrides both previous rules

Two files, with all light rules before the dark rules, forces the latter to have higher precedence. Thus further rules in light can become overly-specific and require more "undoing", if any dark rules affect them.

PR #1049, opened 2025-04-17
Allow override of light/dark mode when auto-detection impossible

lucas-sandery:
Further to #1009, allow setting the correct colour mode for CSS URLs & files specified by `css()`, when auto-detection is impossible or not necessary.


vrana:
The method `css()` now returns URL in the key and `'light'`, `'dark'` or `''` (both) in the value.

PR #1070, opened 2025-05-01
Update functions.inc.php

schucan:
silencing an unnecessary warning

### DIFF
--- a/adminer/include/functions.inc.php
+++ b/adminer/include/functions.inc.php
@@ -685,7 +685,7 @@ function file_open_lock(string $filename) {
 	if (!$fp) {
 		return;
 	}
-	chmod($filename, 0660);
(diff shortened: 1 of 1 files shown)


schucan:
It just happened when I uploaded the file to a web server directory and opened it in my browser. I have been a user of admin for decades by now. This error has never occurred before.

schucan:
Now that I think of it: I had been using adminer in a previous version and was logged in to a database, then replaced a 4.X version of adminer with the latest 5.X version and just reloaded the page and that's when the warning occurred. maybe being logged in was part of the reason why the warning was thrown?

vrana:
What is the exact warning?

schucan:
```
Warning: chmod(): Operation not permitted in /.......php on line 107
```

schucan:
Additional info:
For technical reasons (permission management) I run adminer via a local require_once call.

PR #1081, opened 2025-05-08
Re-added hydra and pepe-linha-dark

PilaScat:
I don't understand why they were removed, they are perfectly working


vrana:
Adminer doesn't depend on them.

PR #1085, opened 2025-05-12
Compile: Fix type of $translations_version.

Daniel-KM:
When minified, the space is removed between the constatn and the number `LANG . 12345678` and `LANG.12345678` is not valid because a float is following the constant.


Daniel-KM:
And there is a [module](https://omeka.org/s/modules/Adminer) that integrates adminer to [Omeka](https://omeka.org/s), so you can add it on your page.

vrana:
You need to update PhpShrink with `git submodule update`.

I've added the link, thanks.

PR #1086, opened 2025-05-14
Plugins pretty-json-column: Encode value to JSON without white spaces

ludekbenedik:
Removing white spaces coming from the form.


vrana:
This function currently does nothing and it seems like that it could be deleted. What are you trying to achieve? It seems that you are just decoding and encoding it.

ludekbenedik:
If an user input is not re-encoded then unnecessary spaces and line breaks are stored in a database if the database uses a text type column for JSON.

The pull request only returns the behavior that existed before (see https://github.com/vrana/adminer/commit/156839142e5c8655689d0aa36748e38b65c0b291).

PR #1090, opened 2025-05-20
Base style fixes + theme update

lucas-sandery:
Moving the RTL rules above allows their precedence from selector specificity, but stops later rules (that should override them) from being undone. Also overriding only constituent properties reduces further conflicts and makes theming easier. Menu open button layout now mirrors layout of LTR, and its rules are now specific to `.js` to repair layout when JS is off/fails to load.

Update to my theme relies on the base style fixes.
Screenshot composite of light and dark modes


vrana:
I've reverted the changes in `adminer.css` that would break something else.

PR #1176, opened 2025-09-24
Update clickhouse.php

Marcus-Nightingale:
Bugfix: table_status($name) in the ClickHouse driver ignored the $name argument and always returned all tables.

This caused Adminer’s UI (e.g. breadcrumbs, links and title) to show the wrong table as selected.

Fixed by applying the $name filter in the query so only the requested table is returned.

### DIFF
--- a/plugins/drivers/clickhouse.php
+++ b/plugins/drivers/clickhouse.php
@@ -272,17 +272,20 @@ if (isset($_GET["clickhouse"])) {
 		return array();
 	}
 
-	function table_status($name = "", $fast = false) {
-		$return = array();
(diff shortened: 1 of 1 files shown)


magnusvin:
You can check the 3 last commits in #1183 on how i fixed. If it is the correct way i am unsure....

Marcus-Nightingale:
> You can check the 3 last commits in #1183 on how i fixed. If it is the correct way i am unsure....

I had a look and there was a few other violations on the phpcs stuff for me. I have created another branch to bring that upto date but we will need @vrana to confirm i think

PR #1199, opened 2025-10-27
Connecting to socket did not work after support for ipv6 in 5.4.0

magnusvin:
image

The commit [f921dafa](https://github.com/vrana/adminer/commit/f921dafa61da0dde151bfddaf0fd27e104fe9078#diff-c82819738ebbfe7ea9821ff520636f41d332c51953a6c828ffa7c2d5a48cfc08R180) breaks behavior that previously worked.

I’ve added comments in the commit messages explaining what each change does and split the commits accordingly.

When the host is empty due to :/unix_socket_path, I now check for that and omit the host in the DSN.

The behavior from [this line](https://github.com/vrana/adminer/commit/f921dafa61da0dde151bfddaf0fd27e104fe9078#diff-c82819738ebbfe7ea9821ff520636f41d332c51953a6c828ffa7c2d5a48cfc08L21) has been moved into the host_port function.

Because 5.4.0 introduced breaking behavior on my part, I manually fixed the adminer.php file like this:
image

Shouldn't the CI tests pass? I agree that I should separate the PRs as suggested in [#1183](https://github.com/vrana/adminer/pull/1183/commits).

### DIFF
--- a/adminer/drivers/mysql.inc.php
+++ b/adminer/drivers/mysql.inc.php
@@ -177,7 +177,7 @@ if (!defined('Adminer\DRIVER')) {
 				}
 				list($host, $port) = host_port($server);
 				return $this->dsn(
-					"mysql:charset=utf8;host=$host" . ($port ? (is_numeric($port) ? ";port=" : ";unix_socket=") . $port : ""),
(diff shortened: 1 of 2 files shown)


magnusvin:
@vrana Can you take a new look on this PR?

vrana:
There's _Undefined array key 3_ in your code. It's also unnecessary complex. There's also a bad `empty()`, Adminer uses `!= ""`.

Issue #1152, opened 2025-09-02
Error in saving empty string in ENUM type column


RadaRD77:
**Adminer version:** 5.3.0
**Compiled:** single file / single language CS
**Driver:** MySQLi only
**Database version:** MySQL 8.0.40 and 8.0.29-21

I have an MySQL table with ENUM column with three values:
"" (yes, this is empty string)
"TYP1"
"TYP2"

I have DEFAULT value for this column as empty string.

When I editing row and trying to save changes, then I have an error "Column 'vypocet' cannot be null". It's because Adminer trying save empty value as NULL instead empty string.

<img width="537" height="336" alt="Image" src="https://github.com/user-attachments/assets/33be1165-db45-4138-82fd-2710e795d3bb" />

<img width="596" height="358" alt="Image" src="https://github.com/user-attachments/assets/80ed4c96-b604-4ce4-98d6-667d92d538c5" />

<img width="412" height="229" alt="Image" src="https://github.com/user-attachments/assets/b380bf15-df0e-42fc-b7f2-a9661203f149" />

Issue #1191, opened 2025-10-17
MySQL / MariaDB - Invalid SQL for ENUM fields


CRCinAU:
**Adminer version:** 5.4.1
**Compiled:** Docker Image
**Driver:** 11.8.3-MariaDB-ubu2404 through PHP extension MySQLi
**Database version:** 11.8.3-MariaDB-ubu2404 through PHP extension MySQLi
**Plugins used:**
AdminerDumpBz2: Dump to Bzip2 format
AdminerEnumOption: Use <select><option> for enum edit instead of <input type="radio">
AdminerTablesFilter: Filter names in tables list
docker\DefaultServerPlugin: Prefills the “Server” field with the ADMINER_DEFAULT_SERVER environment variable.

The following is the DB structure with example data:

<img width="616" height="206" alt="Image" src="https://github.com/user-attachments/assets/53afc75a-e878-4cda-844b-a95fc1f21b01" />

<img width="948" height="300" alt="Image" src="https://github.com/user-attachments/assets/084b6458-6265-4c4f-81f7-53745c7715d7" />

When hitting the Save button, the SQL generated is invalid:

```sql
UPDATE `users` SET
`username` = 'myusername',
`password` = '{SHA512-CRYPT}$6$rou...',
`active` = '',
`group_id` = '0',
`group_admin` = '',
`quota` = '10'
WHERE `username` = 'myusername' AND `username` = 'myusername' COLLATE utf8mb4_bin;
```

PR #1259, opened 2026-03-27
check if "val-" exists in enum value

NReib:
Problem probably introduced in #1152 
Some enum values seem to not be prefixed with "val-" leading to SQL Errors on insertion

### DIFF
--- a/adminer/include/html.inc.php
+++ b/adminer/include/html.inc.php
@@ -317,7 +317,9 @@ function process_input(array $field) {
 		if ($value == "null") {
 			return "NULL";
 		}
-		$value = substr($value, 4); // 4 - strlen("val-")
+		if (preg_match('~val-~', $value)) {
+			$value = substr($value, 4); // 4 - strlen("val-")
+		}
 	}
 	if ($field["auto_increment"] && $value == "") {
 		return null;


NReib:
Just so you know what error this fixes: when adding an entry to a table with an enum, the first 4 letters of that enum were stripped, leading to an sql error in best case

vrana:
How can I reproduce this? Which driver? Are you using some plugins? What do you mean by “adding an entry to a table with an enum”?

NReib:
We use Mariadb / mysqlnd 8.5.4. I'm not aware of any plugins used, but my instance of adminer is shipped with another OSS. Hope this helps.
Set up example table.
```
CREATE TABLE tasks (
    id INT AUTO_INCREMENT PRIMARY KEY,
    title VARCHAR(255) NOT NULL,
    priority ENUM('low', 'medium', 'high') DEFAULT 'medium',
    status ENUM('pending', 'in_progress', 'completed', 'canceled') DEFAULT 'pending',
    created_at TIMESTAMP DEFAULT CURRENT_TIMESTAMP
);
```
Open table in adminer and add new entry
Bildschirmfoto vom 2026-04-30 09-28-16

Bildschirmfoto vom 2026-04-30 09-29-03

Resulting SQL Error - first 4 characters are stripped for enums:
Data truncated for column 'priority' at row 1 09:29:44 SQL-Kommando 🗐
INSERT INTO `tasks` (`title`, `priority`, `status`, `created_at`)
VALUES ('', 'um', 'ing', now());

vrana:
This was fixed by #1191.

NReib:
We are using adminer version 5.4.2, should this fix already be active in that version or only from 5.4.3?

vrana:
This was fixed in 5.4.2 but you also need to update https://www.adminer.org/download/v5.4.2/plugins/enum-option.php.

NReib:
Okay, thank you I will check it out

PR #1261, opened 2026-03-30
Fix blob editing

marcovdb:
- Fix blob editing: blob fields do not appear in `$_POST` so `process_input` returned false


vrana:
Why did you add `if ($file !== null)`?

vrana:
Merged without it.

PR #1285, opened 2026-05-17
feat: prefill login password from query parameter

ismdeep:


vrana:
Passing password in URL is risky as it's logged in several places.

PR #1289, opened 2026-05-31
security: add allowed_classes => false to unserialize() in brute-force protection

XananasX7:
## Summary

`auth.inc.php` reads serialized brute-force tracking data from temp files and passes it to `unserialize()` without specifying `allowed_classes`.

The stored data structure is `array<ip_address, [expiry_time, count]>` — plain arrays of integers. There's no reason for objects here.

Without the restriction, an attacker who can write to the temp directory (e.g., through a path traversal or a symlink attack on the temp file path) can plant a serialized object payload that gets instantiated when Adminer reads its brute-force protection data.

## Fix

Added `['allowed_classes' => false]` to both `unserialize()` calls in `auth.inc.php`. Zero behavior change for valid data.

### DIFF
--- a/adminer/include/auth.inc.php
+++ b/adminer/include/auth.inc.php
@@ -24,7 +24,7 @@ function add_invalid_login(): void {
 	if (!$fp) {
 		return;
 	}
-	$invalids = unserialize(stream_get_contents($fp));
+	$invalids = unserialize(stream_get_contents($fp), ['allowed_classes' => false]);
 	$time = time();
 	if ($invalids) {
 		foreach ($invalids as $ip => $val) {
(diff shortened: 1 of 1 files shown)


XananasX7:
The `build-test` failure is caused by phpcs errors in the CI configuration itself (`ERROR: Referenced sniff "Generic.Functions.CallTimePassByReference" does not exist`) — this is an upstream PHPCS compatibility issue unrelated to the change in this PR. Our change only touches `auth.inc.php` (adding `allowed_classes => false` to one `unserialize()` call), which does not affect the PHP coding standard check. Happy to help fix the CI config if useful.

vrana:
This is problematic only if Adminer is in an environment with some autoloader. In that case, disabling `allowed_classes` doesn't help. I'll use JSON instead.

PR #1301, opened 2026-07-09
PHP 8.5 support: $http_response_header is deprecated

saulery:
Replace $http_response_header which is deprecated by PHP 8.5.

See https://www.php.net/manual/en/reserved.variables.httpresponseheader.php

### DIFF
--- a/plugins/drivers/clickhouse.php
+++ b/plugins/drivers/clickhouse.php
@@ -25,7 +25,7 @@ if (isset($_GET["clickhouse"])) {
 					'max_redirects' => 0,
 				))));
 
-				if ($file === false || !preg_match('~^HTTP/[0-9.]+ 403~i', http_get_last_response_headers()[0]))
+				if ($file === false || preg_match('~^HTTP/[0-9.]+ 403~i', version_compare(PHP_VERSION, '8.4.0', '<') ? $http_response_header[0] : http_get_last_response_headers()[0])) {
 					$this->error = lang('Invalid credentials.');
 					return false;
 				}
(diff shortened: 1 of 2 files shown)


vrana:
`http_get_last_response_headers()` is available since PHP 8.4. `a()[0]` is supported since PHP 5.4. Adminer supports PHP 5.3+.

saulery:
> `http_get_last_response_headers()` is available since PHP 8.4. `a()[0]` is supported since PHP 5.4. Adminer supports PHP 5.3+.

Ok, so I have to add some conditional.

saulery:
Should be fine now.

vrana:
This does not compile in PHP < 5.4. I also prefer `function_exists` over `version_compare`.

saulery:
I have replaced `version_compare `by `function_exists`.

I haven't found a decent online PHP sandbox for testing with PHP 5.3. Do you know of one? How do you handle it, other than installing a full PHP environment on your machine?

vrana:
https://3v4l.org/

saulery:
Should be fine. It's tedious to test since $http_response_header is not generated by https://3v4l.org/ because the sandbox is in CLI mode.

vrana:
The deprecated warning is issued in compile-time so this doesn't fix the problem. Also there are more instances in `clickhouse.php` and `igdb.php` - this should be extracted to a function.

vrana:
Thanks for the effort but I'd rather fix it myself.

saulery:
> Thanks for the effort but I'd rather fix it myself.

Thanks for the fix.

I also spotted other errors related to PHP 8.x that I haven't patched. I'm going to make a list of them.

PR #1309, opened 2026-08-01
before-unload: the page is being left, not reloaded (cs, de)

landsman:
The description of the before-unload plugin says the confirmation appears before the page is *reloaded* in German, which is narrower than what the code does: `onbeforeunload` fires whenever the page goes away — closing the tab, navigating elsewhere, or a reload.

- **de**: "bevor die Seite neu geladen wird" (before the page is reloaded) → "bevor die Seite verlassen wird".
- **cs**: "odnahráním" is not a word Czech uses for this; "opuštěním" is what the Croatian and Polish strings already say ("napuštanja", "rozładowaniem" → leaving).

The English doc-comment, `hr` and `pl` are correct and untouched. `ja` reads "before reloading the page" (再読込み) too, so it may want the same fix, but I do not speak Japanese well enough to propose the wording.

### DIFF
--- a/plugins/before-unload.php
+++ b/plugins/before-unload.php
@@ -33,8 +33,8 @@ onbeforeunload = () => editChanged;
 	}
 
 	protected $translations = array(
-		'cs' => array('' => 'Zobrazí potvrzení před odnahráním stránky, pokud bylo změněno formulářové políčko'),
-		'de' => array('' => 'Zeigt eine Bestätigung an bevor die Seite neu geladen wird, wenn ein Formularfeld geändert wurde'),
+		'cs' => array('' => 'Zobrazí potvrzení před opuštěním stránky, pokud bylo změněno formulářové políčko'),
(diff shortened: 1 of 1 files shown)


landsman:
The red CI here is not from this change. PHPStan only analyses `adminer/` and `editor/` (`conf/phpstan.neon`), so `plugins/before-unload.php` is never looked at, and the same failure reproduces on an unmodified checkout of `main` at 29623c3:

```
Ignored error pattern booleanAnd.alwaysFalse was not matched in reported errors.
```

It arrived with PHPStan 2.2.6 (2.2.5 is green on the same tree) and `php-actions/phpstan@v3` is pinned to `version: latest`.

#1310 fixes it in one line and its own run is green. Happy to rebase this branch once that lands so the check here goes green too.

vrana:
Thanks and sorry about the CI error.

PR #1314, opened 2026-08-03
ClickHouse: Complete driver support

bpbp-boop:
Codex assisted

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -46,6 +46,7 @@
 - MS SQL: Hide table actions and row editing in the sys schema
 - Elasticsearch, ClickHouse: Use default port
 - ClickHouse: Fix nullable columns, fix default values, show server version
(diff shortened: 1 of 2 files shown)


vrana:
Have you extensively tested this with a real ClickHouse instance? Have you read and understood all the generated code?

bpbp-boop:
Yes, we've been using it in-house for a few weeks. The code is quite mechanical, just filling in the functions adminer expects so it works. The main weakness is the lack of index editing or ability to set the primary key during table creation. 

edit: to support primary keys properly for clickhouse (multi column, no auto-increment) would have required bigger changes in Adminer. wasn't sure how far to go with this PR. for our use case reading/editing/exporting data with this driver was enough.

vrana:
Thanks for your contribution. There were some regressions so I've fixed them together with some smaller issues in 9ba86b48. Please test it against Git Adminer (6.0.0-dev in the main branch).

PR #1315, opened 2026-08-03
e2e tests

landsman:
E2E tests verify real user workflows and catch integration regressions that unit tests miss. Their value is even higher in the AI-assisted development era, where code generation is cheap but review and maintenance remain expensive. They provide maintainers with confidence that AI-generated changes preserve existing behavior before merging.

---

By the way, you should probably configure your GitHub Actions so they don't run automatically when someone from outside the repository adds a new workflow. It's a security risk (e.g. cache poisoning and similar attacks).

You can configure this in **Settings → Actions → General**. Under **"Approval for running fork pull request workflows from contributors"**, I'd recommend selecting **"Require approval for all external contributors"**. That prevents workflows from automatically running when someone outside the repository or organization submits a PR, including PRs that modify GitHub Actions workflows.

GitHub docs:
- [Managing GitHub Actions settings for a repository](https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/enabling-features-for-your-repository/managing-github-actions-settings-for-a-repository)
- [Events that trigger workflows](https://docs.github.com/en/actions/reference/workflows-and-actions/events-that-trigger-workflows)

It's also worth avoiding privileged `pull_request_target` workflows unless they're carefully designed, since they run in the context of the base repository and can introduce additional security risks if they execute untrusted code.

### DIFF
--- a/.claude/skills/e2e-tests/SKILL.md
+++ b/.claude/skills/e2e-tests/SKILL.md
@@ -1,0 +1,189 @@
+---
+name: e2e-tests
+description: Write or change Adminer's browser end-to-end tests in tests/e2e/ - Gherkin features run by Behat. Use when adding a scenario or a step, changing the seed data, adding a database driver to the suite, or debugging a failing e2e run.
+---
+
+# Writing Adminer's end-to-end tests
+
+Gherkin scenarios run by Behat, driving Playwright against the dev version and a Testcontainers
+database. [The README](../../../tests/e2e/README.md) has the layout and the commands - read it
+first. This is what writing one has to get right.
+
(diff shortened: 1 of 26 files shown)


vrana:
There are already e2e tests using Katalon, I'd rather keep them. This change also seems very intrusive, it's too much new code I don't like and I don't want.

I don't understand GitHub actions all that well. I just want the checks to run on new PRs without me approving every single run.

vrana:
I've converted the current [e2e tests](https://github.com/vrana/adminer/tree/main/tests) from Katalon to Playwright.

PR #1317, opened 2026-08-06
Update login-password-less.php

niket-agrawal:
removed password_required function causing issues in latest version of adminer


vrana:
The function is defined at https://github.com/vrana/adminer/blob/60e8ff7/adminer/include/auth.inc.php#L73. If you are using non-dev Adminer version then use its version of plugin linked from https://adminer.org/plugins/.

PR #1319, opened 2026-08-10
Designs: Show the table names in the menu of Editor

Daniel-KM:
In editor, five of bundled designs hide every table name in menu. The menu shows a column of identical icons, so the tables cannot be told apart: hever, konya, rmsoft, rmsoft_blue, rmsoft_blue-dark. In fact, the designs style the link to select data as an icon without any text. So just use a more precise css selector.


vrana:
Thank you.

PR #1336, opened 2026-09-15
pev 2 explain

salacr:
I wouldn't lie it's generated by AI.. but i tested and it works great for my use case and i believe that others might like it as well!


vrana:
AI is fine, I use it all the time. The plugin is nice, it just needs some polish:

1. Please define `csp()`:

> Connecting to 'https://cdn.jsdelivr.net/npm/bootstrap@4.5.0/dist/css/bootstrap.min.css.map' violates the following Content Security Policy directive: "connect-src 'self' https://www.adminer.org". The request has been blocked.

> vue.esm-browser.prod.js:1 EvalError: Evaluating a string as JavaScript violates the following Content Security Policy directive because 'unsafe-eval' is not an allowed source of script: script-src 'self' 'unsafe-inline' 'nonce-YmFmNmEyMmQwYzhhN2FiYjlmZWQwZDA3NGU4Yzg1MWY=' 'strict-dynamic'".

Ideally, the plugin wouldn't need `eval` at all. It currently comes from two places: Vue's template compiler (use `vue.runtime.esm-browser.prod.js` instead) and `_.template()`, which PEV2 1.3.0's formatters call. PEV2 1.24.0 no longer calls `_.template()`, so please upgrade (its stylesheet is `dist/pev2.css` and it depends on Bootstrap 5.3).

2. Scrolling the mouse wheel over the modal scrolls the page underneath. Setting `document.documentElement.style.overflow = 'hidden'` while the modal is open (and restoring it on close) should fix it. Also, Esc should close the modal.

3. Mark machine translations with the model name, e.g. `// Claude Opus 5`, and put each translation on a separate line. `Close` isn't translated at all, and `Analyze` is missing in most languages. ANALYZE is an SQL keyword, so don't translate it. The translation key should be the English text itself, which makes the `en` array unnecessary.

4. The Estimate/ANALYZE form posts `query` back to the SQL command page, which runs the whole query again before `EXPLAIN ANALYZE` runs it once more. Please handle this request in `headers()` and return only the plan JSON (see `plugins/sql-gemini.php`). Also, run `EXPLAIN ANALYZE` inside `BEGIN` … `ROLLBACK`, because it actually executes the query, including any functions with side effects.

5. Compare `Adminer\DRIVER` directly without `defined()`, print the `<style>` only once, and don't rely on `previousElementSibling.previousElementSibling`.

6. The `@link` points to a fork. Please link to your published plugin and to dalibo/pev2, and add `@author`.

Then publish it in your repo or on Gist and let me know here.

PR #1338, opened 2026-09-17
Plugins: Translate to zh

Hui-1121:
The `zh` (简体中文) column for plugins in the leaderboard was at 6% – only `select-email.php` had the translations. This fills in the remaining 104 messages across 46 plugins (47 descriptions and 57 UI strings), which brings `zh` to 100% in that column.

Notes on the translations:

- Terminology follows the existing core `adminer/lang/zh.inc.php` so that the plugins read the same as Adminer itself, e.g. `Collation` → 校对, `New item` → 新建数据, `Auto Increment` → 自动增量, `original` → 原始.
- The three messages with `%d` are plain strings instead of arrays because Chinese has no plural inflection – `lang.php` reports `identical plural forms` for duplicated forms, and the `zh` entry already present in `select-email.php` does the same.
- `dump-xml.php` keeps the literal tags of the English description and translates only `value`, like the other languages do.

Checks:

- `php lang.php zh` reports no placeholder or full stop problems and running it again changes nothing.
- `php lang.php stats` updated the table in `adminer/lang/README.md`.
- `composer check` passes (phpcs and phpstan; no JavaScript was touched).
- `Plugins::checksum()` is identical before and after for all 47 files, so no plugin gets reported as modified.
- Rendered on the development server against MySQL 8.0 with 37 of these plugins loaded: the descriptions in `Loaded plugins` and the strings of `config`, `menu-links`, `timeout`, `table-structure`, `table-indexes-structure`, `tables-filter`, `import-csv` (including `Table has been created.`, `%d row(s) have been imported.`, `Column %s does not exist.` and `File must be in UTF-8 encoding.` from a real import), `enum-option`, `select-email`, `row-numbers`, `backward-keys` and `sql-gemini` all show up in Chinese, with no PHP errors on any of the pages. The login page strings of `login-passkey` and `login-otp` were checked with those two plugins loaded separately.

Not exercised in a running Adminer, only through `Plugin::lang()`: `%d e-mail(s) have been sent.` (needs a working MTA), `%d byte(s)` in `select-image` (needs image data), `Just a sec…` in `sql-gemini` (needs an API key) and `Only some plugins support configuration, e.g. %s.` (the source marks that branch as currently unused).


vrana:
This is great work, thanks. Why have you closed the PR? Can I still use it?

vrana:
I've merged it with the `Claude Opus 5` marker. If you confirm that you checked all the translations then I can remove the marker.

